PR #2, opened 2024-01-13
Country and year setters

SimonJnsson:
Currently, the setters for year and country in the Holidays class will actually create a new instance of the class. This results in chaining of both methods not working. This PR fixes that issue by simply setting the property and returning the current instance. It also updates the tests that are currently broken because of this issue.


Nielsvanpach:
Thanks for your PR, Simon. I've made some API changes in the meantime.

PR #5, opened 2024-01-17
Danish holidays

RootAccessPlease:
Added the public holidays for Denmark.


Nielsvanpach:
You should run the tests, so a snapshot of the dates can be generated.

tests/.pest/snapshots/DenmarkTest/it_can_calculate_danish_holidays.snap
```diff
@@ -0,0 +1,46 @@
...
+        "name": "Pinse",
+        "date": "2024-05-19"
+    },
+    {
+        "name": "Anden Pinsedag",
+        "date": "2024-05-20"
+    },
+    {
+        "name": "Juleaften",
+        "date": "2025-12-12"
```
Nielsvanpach, line 36: Could you verify all generated results? It seems like something is not right yet.

steffen25:
@RootAccessPlease Nice PR!

Do you think it makes sense to add support for Great Prayer Day prior to 2024?
Maybe it is enough to mention that in the docs somehow.

tests/.pest/snapshots/DenmarkTest/it_can_calculate_danish_holidays.snap
RootAccessPlease: Sorry! Should be correct now.

RootAccessPlease:
@steffen25 Great point! I've taken that into account in the latest commit.

Nielsvanpach:
Thanks!

Tor2r:
While making the holidays for Norway, I started by copying the Danish ones. Found an error with Easter Monday / Anden Påskedag
Ref: https://publicholidays.dk/2024-dates/

PR #6, opened 2024-01-17
Implement holidays for Austria

felixfrey:


src/Countries/Austria.php
```diff
@@ -0,0 +1,43 @@
...
+        return array_merge([
+            'Neujahr' => '01-01',
+            'Heilige Drei Könige' => '01-06',
+            'Staatsfeiertag' => '05-01',
+            'Mariä Himmelfahrt' => '08-15',
+            'Nationalfeiertag' => '10-26',
+            'Allerheiligen' => '11-01',
+            'Mariä Empfängnis' => '12-08',
+            'Christtag' => '12-25',
+            'Stefanitag' => '12-25',
```
Nielsvanpach, line 26: Should't this be 26th? https://publicholidays.at/2024-dates/

Nielsvanpach:
Thanks!

marlonbasten:
I added state-specific holidays in Austria to my PR. Can you maybe check if those are correct as I am not Austrian? 😅 #37

felixfrey:
> I added state-specific holidays in Austria to my PR. Can you maybe check if those are correct as I am not Austrian? 😅 #37

The thing with state-specify holidays in Austria ist that they are not real "holidays" - shops are open and employees have to work. State-specifiy holidays mostly only apply to universities, schools and sometimes to government agencies. It is complicated. (https://de.wikipedia.org/wiki/Feiertage_in_%C3%96sterreich). The question is what the definition of a holiday is for this package.

Nielsvanpach:
I made this package to calculate the official public holidays. If the state-specific holidays in Austria are not real holidays, and people still have to work, I wouldn't include them.

PR #7, opened 2024-01-17
Use more common names for Dutch holidays

CasEbb:
Use the names of holidays as they are known in the Netherlands


Nielsvanpach:
Thanks!

PR #8, opened 2024-01-17
Adding German holidays

arnebr:
This uses the nation-wide holidays for Germany from `https://de.wikipedia.org/wiki/Gesetzliche_Feiertage_in_Deutschland`

Regional state holidays are not part implemented in the library. Maybe @freekmurze  as future plan?


arnebr:
For easy implementing we could use something like de-sh but that's not "normed".
For Austria it would be the same. We could look at libraries like: https://github.com/umulmrum/holiday

arnebr:
### Notes on German Holidays 


- Easter Sunday/Ostersonntag and Whit Sunday/Pfingstsonntag are not public holidays in most states.
- Every Sunday is a public holiday in Hesse.
- Assumption/Mariä Himmelfahrt is a public holiday in about 1700 of about 2000 communities in Bavaria. This is 
  implemented as a partial holiday, so you might want to add your logic that filters this date.
- In Bavaria schools are closed on Assumption Day and on Repentance and Prayer Day. There is no special handling of
  this "behavior".
- Corpus Christi/Fronleichnam is a public holiday in some communities in Saxony and Thuringia.
- Regional holidays with traditionally (but not publicly) limited opening hours are not considered yet (e.g. carnival 
  days).
- Some ports at the German Ocean celebrate some holidays as "High Holidays". It is generally not allowed to work on 
  these days, and work time ends at 12 o'clock the day before. This is not considered yet.
- Also there are some quiet days that are not public holidays in various states. This is also not considered yet.
- New states of Germany are treated as if they had always been part of the Federal Republic of Germany. For years
  before 1990 holidays are therefore not correct.

source: https://github.com/umulmrum/holiday

Nielsvanpach:
> Regional state holidays are not part implemented in the library. Maybe freekmurze as future plan?

We are open to support this, feel free to open a PR.
Maybe the Country class can be extended with a region property, which could be used in `allHolidays()` ?

Nielsvanpach:

Could you get this PR up to date?

arnebr:
Will do 😊

arnebr:
@Nielsvanpach I think #37 superseeds this PR, since it's implemented all states 😊

Nielsvanpach:
I'll merge this PR for now. We are using a simpler variant for implementing regions, departments, states, ...

marlonbasten:
@arnebr As my PR won't get merged, there's still a need for German state holidays. 😕 Feel free to add this to your PR!

arnebr:
Maybe @marlonbasten you could do a review of the holidays :)  I implemented all of them now. 
@Nielsvanpach hope that´s the way you would like to see it.

tests/Countries/GermanyTest.php
```diff
@@ -0,0 +1,80 @@
...
+    expect($holidays)
+        ->toBeArray()
+        ->not()->toBeEmpty();
+
+    expect(count($holidays))->toBe($totalHolidays);
+
+    expect(formatDates($holidays))->toMatchSnapshot();
+})->with(
+    [
+        ['BW',12],
```
Nielsvanpach, line 64: Have you actually checked the results in the snapshots? Otherwise having only 1 region test is fine.

Nielsvanpach (commented):
PR looks good, thanks!

tests/Countries/GermanyTest.php
Nielsvanpach: Great, thanks for this!

Nielsvanpach (approved):
PR looks great, thanks for your work! 👍

PR #9, opened 2024-01-17
Adds support for Portuguese Holidays

lamelas:


tests/.pest/snapshots/Countries/PortugalTest/it_can_calculate_portuguese_holidays.snap
```diff
@@ -0,0 +1,58 @@
+[
+    {
+        "name": "Dia de Ano Novo",
+        "date": "2024-01-01"
+    },
+    {
+        "name": "Carnaval",
```
Nielsvanpach, line 7: Is this an official public holiday? It's not on the list here: https://publicholidays.pt/2024-dates/

lamelas: That’s a tricky one, actually. It is not an official mandatory holiday (hence not appearing on the list) but it is a facultative holiday so it appears on most work contracts and the government validates it every year as a holiday for civil servants. In practical terms it is a holiday since nobody works and everything is closed.

joaoffm: Unfortunately like you said it’s not oficial, many companies ignore it.
It should be omitted from this list, in my opinion.
Good job with the PR 💪

lamelas: Fair enough. I'll remove it then.

freekmurze:
Thanks!

PR #12, opened 2024-01-17
Indian National Holidays

RohanSakhale:
This PR introduces national holidays in India referenced from [Wiki Public Holidays Page](https://en.wikipedia.org/wiki/Public_holidays_in_India)

### New Changes

1. Introduced India as a country
2. Added test case for the Indian Holidays to be verified

## To ask

Many times there are state-specific holidays in many countries, we can optionally introduce state-specific holidays that can be passed as active state variables or variables when checking `isHoliday` this would make the repository very useful when utilizing it across countries.


Nielsvanpach:
We are open to support states or regions.

As this is also discussed on the [German PR](https://github.com/spatie/holidays/pull/8), I've added it as feature request: https://github.com/spatie/holidays/discussions/14

RohanSakhale:
@Nielsvanpach attempted to add region support, can you review

lokesh-zersys:
@RohanSakhale Most of the holidays you have added will change every year,  eg Diwali will be on a different date each year.
So you cannot add such holidays. Unless we come up with some logic to predict the date.

RohanSakhale:
@lokesh-zersys yes that is true, that thought was in my mind, but need to think of a way out to provide dates based on Hindu Calendar, which I was searching but couldn't found concreate way out unless we start defining them ourself in the repo

jigarakatidus:
In my opinion, we should follow **ONLY** our Country's and respective State's(Region) Law. 
For those who won't to change the list of Holiday's, should extend the India class and override the methods as required. Yes, including Diwali, Christmas and Eid.

For e.g. according to Maharastra Law. There are only 4.

**3 National Holiday**
- 26th Jan - Republic Day
- 15th Aug - Independence Day
- 2nd Oct - Gandhi Jayanti

**1 State Holiday**
- 1st May - Labour Day / Maharashtra Day

jigarakatidus:
Also, list of holidays changes based on sector as well, not just region. E.g. banking and school sector will have more holidays than offices. Just a thought to keep in mind when creating a list of Holidays that is generic for all.

Nielsvanpach:

Could you get this PR up to date or maybe better, start a new one?

RohanSakhale:
@Nielsvanpach updated based on documented structure

I believe we should also support `region` when creating country object using `Holidays::for`, please suggest your views

src/Countries/India.php
```diff
@@ -0,0 +1,475 @@
...
+            'CG', 'Chhattisgarh' => [
+                'Makar Sankranti' => '01-14',
+                'Vasant Panchami' => '02-16',
+                'Maha Shivaratri' => '03-11',
+                'Holi' => '03-29',
+                'Good Friday' => '04-02',
+                'Ram Navami' => '04-21',
+                'Mahavir Jayanti' => '04-25',
+                'Buddha Purnima' => '05-26',
+                'Eid al-Fitr' => '05-14',
```
Nielsvanpach, line 113: I don't think a fixed list can be used for Islamic holidays

Nielsvanpach:
Thanks, but I'm closing this PR as it doesn't feel complete yet. Feel free to reopen once fixed!

src/Countries/India.php
RohanSakhale: There we would need a manual intervention for adding Islamic and Hindu holidays based on Moon based calendar, this can either come through config or via class abstract methods

PR #15, opened 2024-01-17
Add French holidays

levrailoup:


wblondel:
Hi @levrailoup 
According to [this page](https://travail-emploi.gouv.fr/droit-du-travail/les-conges-payes-et-les-conges-pour-projets-pro-et-perso/article/les-jours-feries-et-les-ponts), these holidays also need to be included:

- lundi de Pâques
- Ascension
- lundi de Pentecôte

Also, in the departments of Moselle, Bas-Rhin and Haut-Rhin, the list is slightly different: https://www.legifrance.gouv.fr/codes/article_lc/LEGIARTI000006902635

In Guadeloupe, Guyane, Martinique, Mayotte, La Réunion, Saint-Barthélemy and Saint-Martin, which are all part of France, there are also additional holidays: https://www.legifrance.gouv.fr/codes/article_lc/LEGIARTI000035902463

I am not sure if this package handles regional differences.

levrailoup:
Lundi de Pentecôte, Ascension and Lundi de Pâques are already included in the PR : As they are dependent upon Easter date, they can be found in France::variableHolidays().

As for the other dates you mention... I'm not quite sure what is expected from this package : should we have a specific class for each region ? Country codes don't include regions AFAIK, so it might prove difficult. "Vendredi Saint" is even town-specific...

In the meantime, the current PR includes all common holidays for mainland France.

src/Countries/France.php
```diff
@@ -0,0 +1,41 @@
...
+            'Armistice 1918' => '11-11',
+            'Noël' => '12-25',
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        $easter = CarbonImmutable::createFromTimestamp(easter_date($year))
+            ->setTimezone('Europe/Brussels');
```
Kocal, line 33: Why not using the Paris timezone?

```suggestion
            ->setTimezone('Europe/Paris');
```

levrailoup: You are absolutely right

src/Countries/France.php
```diff
@@ -0,0 +1,41 @@
...
+use Carbon\CarbonImmutable;
+
+class France extends Country
+{
+    public function countryCode(): string
+    {
+        return 'fr';
+    }
+
+    /** @return array<string, CarbonImmutable> */
```
stof, line 14: this return type is wrong (but anyway, you should not need to specify it here. Inheriting it from the parent is better)

Nielsvanpach:

wblondel:
> Lundi de Pentecôte, Ascension and Lundi de Pâques are already included in the PR : As they are dependent upon Easter date, they can be found in France::variableHolidays().

My bad. I must have forgotten to open my eyes 😅

Nielsvanpach:
Could you rebase this branch, I think it should allow the tests to run here

Nielsvanpach:
This PR looks good! I'll wait for the department based holidays before merging this.

I've added documentation on how I think we should manage regions, states or departments:

levrailoup:
Added support for region-specific holidays. It uses [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2:FR) codes for now, but I am open to suggestions.

Nielsvanpach:
Thanks for your work on this one!

PR #17, opened 2024-01-17
Add Malaysia holiday

pisyek:


Nielsvanpach:
Could you rebase this branch, in order to run the tests here?

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    /**
+     * @param int $year
+     * @return array<string, CarbonImmutable>
+     */
+    protected function fixHolidays(int $year): array
+    {
+        return [
+            'Hari Pekerja' => CarbonImmutable::createFromDate($year, 5, 1, $this->timezone),
```
Nielsvanpach, line 41: Could you use the format like other Country classes do? The timezone part can be removed.

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,535 @@
...
+            'phg' => $this->regionPahang($year),
+            'png' => $this->regionPenang($year),
+            'prk' => $this->regionPerak($year),
+            'pls' => $this->regionPerlis($year),
+            'pjy' => $this->regionPutrajaya($year),
+            'sbh' => $this->regionSabah($year),
+            'swk' => $this->regionSarawak($year),
+            'sgr' => $this->regionSelangor($year),
+            'trg' => $this->regionTerengganu($year),
+            default => array_merge(
```
Nielsvanpach, line 99: By default I wouldn't return any region specific holidays. What you think?

tests/Countries/MalaysiaTest.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Spatie\Holidays\Tests\Countries;
+
+use Carbon\CarbonImmutable;
+use Spatie\Holidays\Holidays;
+
+it('can calculate Malaysia holidays', function () {
```
Nielsvanpach, line 8: can you add a specific test for a region too?

src/Countries/Malaysia.php
pisyek: i think it is better to add check if region is set; then throw a new exception invalid region if no match. better flow right? @Nielsvanpach

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,562 @@
...
+    {
+        return 'my';
+    }
+
+    protected function allHolidays(int $year): array
+    {
+        return array_merge(
+            $this->fixHolidays($year),
+            $this->variableHolidays($year)
+        );
```
Nielsvanpach, line 50: I think I'd prefer regions holidays in a dedicated method here

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,562 @@
...
+    }
+
+    /**
+     * @param int $year
+     * @return array<string, string>
+     */
+    protected function fixHolidays(int $year): array
```
Nielsvanpach, line 57: Could you format this like other countries do? The fixed dates are not in a separate method, but just in the allHolidays() method

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,562 @@
...
+        ];
+    }
+
+    /**
+     * @param int $year
+     * @return array<string, CarbonImmutable|string>
+     */
+    protected function variableHolidays(int $year): array
+    {
+        $variableHolidays = [
```
Nielsvanpach, line 73: It's hard to get a grip of this. Could you format this in a more readable way? You can group the chinese related holidays together in a dedicated method, the same for the islamic related holidays.

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,562 @@
...
+    }
+
+    /**
+     * @param int $year
+     * @return array<string, CarbonImmutable|string>
+     */
+    protected function regionPahang(int $year): array
+    {
+        return [
+            'Tahun Baru' => CarbonImmutable::createFromDate($year, 1, 1, $this->timezone),
```
Nielsvanpach, line 300: A lot of holidays seem to be duplicated across regions.

Nielsvanpach (changes requested):

Nielsvanpach:
I'm probably merging this PR soon, so might be best to use his implementation of the ChineseCalendar: https://github.com/spatie/holidays/pull/115

pisyek:
> I'm probably merging this PR soon, so might be best to use his implementation of the ChineseCalendar: #115

i'll update my PR once #115 merge.

tests/Countries/MalaysiaTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+    expect($holidays)
+        ->toBeArray()
+        ->not()->toBeEmpty();
+
+    expect(count($holidays))->toBe($totalHolidays);
+
+    expect(formatDates($holidays))->toMatchSnapshot();
+})->with([
+    ['jhr', 18],
```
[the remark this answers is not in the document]
pisyek: yes, to date it is correct.

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,540 @@
...
+        };
+    }
+
+    /**
+     * @return array<string, CarbonImmutable|string>
+     */
+    protected function regionJohor(int $year): array
+    {
+        $johorHolidays = [
+            'Hari Keputeraan Sultan Johor' => CarbonImmutable::createFromDate($year, 3, 23, $this->timezone),
```
Nielsvanpach, line 151: Shouldn't be needed to use timezones here.

Nielsvanpach:
```suggestion
            'Hari Keputeraan Sultan Johor' => '03-23',
```

Nielsvanpach: I prefer this format, if possible

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,540 @@
...
+            'Hari Thaipusam' => $this->hariThaipusam($year),
+            'Awal Ramadan' => $this->awalRamadan($year),
+            'Hari Deepavali' => $this->hariDeepavali($year),
+        ];
+
+        $hariHolJohor = match ($year) {
+            2020 => CarbonImmutable::createFromDate($year, 9, 24, $this->timezone),
```
Nielsvanpach, line 158: Timezones shouldn't be needed (this is for all the CarbonImmutable objects using the timezone.

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,540 @@
...
+            'Hari Ulang Tahun Pertabalan Sultan Terengganu' => CarbonImmutable::createFromDate($year, 3, 4, $this->timezone),
+            'Hari Nurul Al-Quran' => $this->nuzulQuran($year),
+            'Hari Keputeraan Sultan Terengganu' => CarbonImmutable::createFromDate($year, 4, 26, $this->timezone),
+            'Hari Arafah' => $this->hariArafah($year),
+            'Hari Raya Aidiladha Hari Kedua' => $this->hariAidiladha($year)->addDays(1),
+            'Hari Deepavali' => $this->hariDeepavali($year),
+        ];
+    }
+
+    private function newYear(): string
```
Nielsvanpach, line 415: No need for this separate function, you can just use `'01-01'`

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,540 @@
...
+    {
+        return '01-01';
+    }
+
+    private function lunarNewYear(int $year): CarbonImmutable
+    {
+        return $this->chineseToGregorianDate('01-01', $year);
+    }
+
+    protected function getIslamicFormatter(): IntlDateFormatter
```
Nielsvanpach, line 425: If possible, I prefer this in a separate trait, just like the ChineseCalender.php

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,540 @@
...
+    {
+        return $this->islamicCalendar('01/10', $year);
+    }
+
+    protected function hariWilayah(int $year): CarbonImmutable
+    {
+        return CarbonImmutable::createFromDate($year, 2, 1, $this->timezone);
+    }
+
+    protected function israkMikraj(int $year): CarbonImmutable
```
Nielsvanpach, line 471: I don't really like all this separate functions. You can just use `$this->islamicCalendar();` instead

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,540 @@
...
+    }
+
+    protected function nuzulQuran(int $year): CarbonImmutable
+    {
+        return $this->islamicCalendar('17/9', $year);
+    }
+
+    protected function hariKeputeraanYDP(int $year): ?CarbonImmutable
+    {
+        return match (true) {
```
Nielsvanpach, line 500: This one you can keep, as there is additional logic needed

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,540 @@
...
+    {
+        return $this->islamicCalendar('1/1', $year, true);
+    }
+
+    protected function maulidurRasul(int $year): CarbonImmutable
+    {
+        return $this->islamicCalendar('12/3', $year, true);
+    }
+
+    protected function hariWesak(int $year): string
```
Nielsvanpach, line 526: No need for a specific function

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,540 @@
...
+    {
+        return '01-01';
+    }
+
+    protected function hariThaipusam(int $year): string
```
Nielsvanpach, line 531: No need for a specific function

src/Countries/Malaysia.php
```diff
@@ -0,0 +1,540 @@
...
+    {
+        return '01-01';
+    }
+
+    protected function hariDeepavali(int $year): string
```
Nielsvanpach, line 536: No need for a specific function

tests/Countries/MalaysiaTest.php
```diff
@@ -0,0 +1,69 @@
...
+    $holidays = Holidays::for(Malaysia::make($region))->get();
+
+    expect($holidays)
+        ->toBeArray()
+        ->not()->toBeEmpty();
+
+    expect(formatDates($holidays))->toMatchSnapshot();
+})->with([
+    ['jhr', 18],
+    ['kdh', 19],
```
Nielsvanpach, line 34: totalHolidays seems unused, you can remove that part.

Nielsvanpach (changes requested):
Perhaps you could take a look at other classes, and refactor the code to make it look more like other Countries?

Examples: Belgium, Albania, Jamaica, ..

Nielsvanpach:
Thanks for all the work!

PR #19, opened 2024-01-17
Add Holidays - Pakistan

muhammad-nauman:
### Overview:
This pull request adds the country class for Pakistan. It lists the fixed federal holidays from the list [here](https://en.wikipedia.org/wiki/Public_holidays_in_Pakistan).

All the relevant test files are also added.

### DIFF
--- a/src/Countries/Pakistan.php
+++ b/src/Countries/Pakistan.php
@@ -11,10 +11,6 @@ class Pakistan extends Country
         return 'pk';
     }
 
-    /**
-     * @return array<string, CarbonImmutable>
-     * @link https://en.wikipedia.org/wiki/Public_holidays_in_Pakistan
-     */
     protected function allHolidays(int $year): array
     {
         return array_merge([
--- a/src/Countries/Pakistan.php
+++ b/src/Countries/Pakistan.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\CarbonImmutable;
+
+class Pakistan extends Country
+{
+    public function countryCode(): string
+    {
+        return 'pk';
+    }
+
+    /**
+     * @return array<string, CarbonImmutable>
+     * @link https://en.wikipedia.org/wiki/Public_holidays_in_Pakistan
+     */
+    protected function allHolidays(int $year): array
+    {
+        return array_merge([
+            'Kashmir Solidarity Day' => '02-05',
+            'Pakistan Day' => '03-23',
+            'Labour Day' => '05-01',
+            'Independence Day' => '08-14',
+            'Iqbal Day' => '11-09',
+            'Quaid-e-Azam Day' => '12-25',
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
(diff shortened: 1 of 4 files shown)


src/Countries/Pakistan.php
```diff
@@ -0,0 +1,36 @@
...
+
+class Pakistan extends Country
+{
+    public function countryCode(): string
+    {
+        return 'pk';
+    }
+
+    /**
+     * @return array<string, CarbonImmutable>
```
Nielsvanpach, line 15: No need for the return type or the @link type

Nielsvanpach (commented):
Thanks! Could you rebase this branch, in order to run the tests here?

src/Countries/Pakistan.php
muhammad-nauman: Removed the doc block.

muhammad-nauman:
@Nielsvanpach I do not understand the reason behind the failing action, can you help me with that?

src/Countries/Pakistan.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    public function countryCode(): string
+    {
+        return 'pk';
+    }
+
+    protected function allHolidays(int $year): array
+    {
+        return array_merge([
+            'Kashmir Solidarity Day' => '02-05',
```
Nielsvanpach, line 17: You can also add translations: 
https://github.com/spatie/holidays/tree/main/lang

muhammad-nauman: Added

src/Countries/Pakistan.php
```diff
@@ -0,0 +1,32 @@
...
+            'Independence Day' => '08-14',
+            'Iqbal Day' => '11-09',
+            'Quaid-e-Azam Day' => '12-25',
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        // The variable holidays are all follow lunar calendar so their dates are not confirmed.
```
Nielsvanpach, line 29: You can look at how Turkey solved this:

tests/Countries/PakistanTest.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Spatie\Holidays\Tests\Countries;
+
+use Carbon\CarbonImmutable;
+use Spatie\Holidays\Holidays;
+
+it('can calculate pakistan holidays', function () {
+    CarbonImmutable::setTestNowAndTimezone('2024-01-01');
```
Nielsvanpach, line 9:
```suggestion
    CarbonImmutable::setTestNow('2024-01-01');
```

src/Countries/Pakistan.php
muhammad-nauman: I checked the implementation, the solution provides the dates but there is no surety that the calendar will follow thsoe exact dates. Should I go for it anyway?

muhammad-nauman: @Nielsvanpach I tried adapting the solution. The problem is, that we have 4 different sets of holidays that follow the lunar calendar(total of 10 holidays) and it is making the code a lot more complex and even after that, the dates will not be accurate.

PR #26, opened 2024-01-17
Add United States Holidays

arcdigital:
I've added all US Bank holidays.

### DIFF
--- a/src/Countries/UnitedStates.php
+++ b/src/Countries/UnitedStates.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\CarbonImmutable;
+
+class UnitedStates extends Country
+{
+    public function countryCode(): string
+    {
+        return 'us';
+    }
+
+    protected function allHolidays(int $year): array
+    {
+        $holidays = array_merge([
+            'New Year\'s Day' => '01-01',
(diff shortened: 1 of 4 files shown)


arcdigital:
I've updated with all the bank holidays. The US is tricky but I think that makes the most sense.

src/Countries/UnitedStates.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        return 'us';
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function allHolidays(int $year): array
+    {
+        return array_merge([
+            'New Year\'s Day' => '01-01',
+            'Juneteenth National Independence Day' => '06-19',
```
Nielsvanpach, line 19: Am I right this is a only recent official holiday? For historical correctness, could you not include this for the years it wasn't an official holiday?

Nielsvanpach (commented):
Thanks! Could you rebase your branch with our main branch? That way we can run the tests here.

Also, are there region based holidays in the US, which need to be taken care of? I've added notes in the readme for this.

arcdigital:
Sure I can do that, yes it is a newer holiday. I'll add that logic in.

arcdigital:
Just rebased it. I'll add the logic for Juneteenth.

Nielsvanpach:
Thanks!

PR #27, opened 2024-01-17
Add Romanian Holidays

florinp:


acip:
Excellent! 
A few adjustments:

Starting **2024** there are two more dates:
* Jan 6th
* Jan 7th

It may be worth implementing historical data:
* Adormirea Maicii Domnului since 2018
* Sfântul Andrei since 2012
* Ziua copilului since 2017
* Ziua Unirii Principatelor Române since 2016

Sources: [Sărbători publice în România](https://ro.wikipedia.org/wiki/Adormirea_Maicii_Domnului) and English version: https://en.wikipedia.org/wiki/Public_holidays_in_Romania

Official records might also help

msucevan (approved):

src/Countries/Romania.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    protected function allHolidays(int $year): array
+    {
+        return array_merge([
+            'Anul Nou' => '01-01',
+            'A doua zi de Anul Nou' => '01-02',
```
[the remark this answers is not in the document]
florinp: I already added with a second commit

src/Countries/Romania.php
```diff
@@ -0,0 +1,78 @@
...
+        $c = $year % 19;
+        $d = (19 * $c + 15) % 30;
+        $e = (2 * $a + 4 * $b - $d + 34) % 7;
+        $month = floor(($d + $e + 114) / 31);
+        $day = (($d + $e + 114) % 31) + 1;
+
+        $gregorianEaster = CarbonImmutable::createFromDate($year, $month, $day);
+
+        // Orthodox Church uses Julian calendar to calculate Easter
+        return $gregorianEaster->addDays(13);
```
alexmanase, line 69: https://www.php.net/manual/en/function.easter-date.php#121559

> 13 days is not a constant. It's an accumulated error fixed in Gregorian and should be calculated with this formula: (int)($year / 100) - (int)($year / 400) - 2

florinp: I added the formula. I missed that one, thanks for the info.

src/Countries/Romania.php
```diff
@@ -0,0 +1,81 @@
...
+
+class Romania extends Country
+{
+    public function countryCode(): string
+    {
+        return 'ro';
+    }
+
+    /**
+     * {@inheritDoc}
```
Nielsvanpach, line 15: No need for the docblocks here

src/Countries/Romania.php
```diff
@@ -0,0 +1,81 @@
...
+        return [
+            'Vinerea Mare' => $goodFriday,
+            'Paștele' => $easter,
+            'A doua zi de Paște' => $easterMonday,
+            'Rusaliile' => $pentecost,
+            'A doua zi de Rusalii' => $pentecostMonday,
+        ];
+    }
+
+    protected function orthodoxEaster(int $year): CarbonImmutable
```
Nielsvanpach, line 56: Where does this code come from? I would like to understand things like this before merging.

Nielsvanpach (commented):
Thanks! Could you rebase this branch on our repo's main branch? That way we can run the tests here

src/Countries/Romania.php
florinp: Source: https://www.php.net/manual/en/function.easter-date.php#83794
Theory: https://www.algorithm-archive.org/contents/computus/computus.html

florinp: I updated the code to use the function from Country class.

Nielsvanpach:
Thanks!

PR #30, opened 2024-01-17
Add holidays for South Africa

pierredup:


Nielsvanpach:
Thanks! Could you rebase on main, that way I can run the tests here

src/Countries/SouthAfrica.php
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+
+        return array_merge($holidays, $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        $easter = CarbonImmutable::createFromTimestamp(easter_date($year))
```
Nielsvanpach, line 56: Could you use `$this->easter($year)` instead?

tests/Countries/SouthAfricaTest.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Spatie\Holidays\Tests\Countries;
+
+use Carbon\CarbonImmutable;
+use Spatie\Holidays\Holidays;
+
+it('can calculate south africa holidays', function () {
+    CarbonImmutable::setTestNowAndTimezone('2022-01-01');
```
Nielsvanpach, line 9: Could you use 2024? That's easier to validate for correctness.

Nielsvanpach:
There are 2 South Africa PRs, but with different results it seems. Could you both verify which one is correct? https://github.com/spatie/holidays/pull/32

pierredup:
Both PRs looks to implement the same days. This one handles an additional edge case, where if a holiday rolls over from the Sunday to the Monday, it shouldn't be added if the Monday is already a Holiday (E.G if 25 Dec falls on a Sunday, it can't roll over to the Monday since the 26 Dec (the following Monday) will already be a holiday).

johanmeiring:
Ah yes, the edge case that is covered here is indeed correct.  This is the more complete implementation, and should be the one that gets merged.  I'll close my PR :-)

P.S. I'll try to remember to open a PR once we know what our general election date will be, as that is very likely to be proclaimed a public holiday as well.

PR #32, opened 2024-01-17
Add South African holidays

johanmeiring:
This PR adds South African public holidays, taking into account the law which states that if a public holiday lands on a Sunday then it is also observed the following Monday.

### DIFF
--- a/src/Countries/SouthAfrica.php
+++ b/src/Countries/SouthAfrica.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\CarbonImmutable;
+
+class SouthAfrica extends Country
+{
+
+    public function countryCode(): string
+    {
+        return 'za';
+    }
+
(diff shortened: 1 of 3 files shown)


Nielsvanpach:
Thanks! Could you rebase on main, so I can run the tests?

Also, there seems to be 2 PRs for South Africa. Might be best if you worked together on 1 PR: https://github.com/spatie/holidays/pull/32

johanmeiring:
Hi @Nielsvanpach !  I've rebased as requested.  As for the other PR... what a coincidence, and implementations seem to be nearly identical.  How would you propose we both work on the same PR?  Or I suppose you could just merge whichever one ends up being ready first or looks best to you?  (I don't mind at all either way)

JonathanAspeling:
Yeah! South Africa represented! Nice one @johanmeiring 🎉🎉🎉🎉🎉💪💪💪

johanmeiring:
Closing in favour of #30

PR #34, opened 2024-01-17
Add Spanish Holidays

jaltez:


markitosgv:
Hello, first thanks for the PR, but there are some things that need to be fixed:

- In Spain, each autonomous community defines 4 days that do not have to coincide between them.
- Each municipality also sets 2 days (this could be excluded from the scope of this library).

Then there is an issue and it is that national holidays do not always fall on the same days, sometimes the 1st of Jan if it falls on a Sunday is moved to the 2nd day or something like that.

How do you think this issue could be handled, or should there really be a PR each time the holidays for the following year are approved?

Thank you!

jaltez:
I think it is a problem that will occur in multiple locations, while the scope of the library only reaches countries and not country regions. I have only included national holidays that are common to all autonomous communities.

It would be necessary to be able to go down a level to filter by autonomous community or obtain all of them, indicating which community it belongs to (similar to the behavior of Google Calendar).

Regarding the free placement of holidays, perhaps it would be necessary to be able to specify it by year, and yes, update periodically after approval.

What do you think?

Thanks!

Nielsvanpach:
Could you rebase your branch on our main branch? That way I can run the tests

jaltez:
Done.

Nielsvanpach:
It seems like Spain has some additional holidays, based on the regions? We added some docs on how we think this could be implemented. Could you extend this PR with support? Thanks!

jaltez:
Yes, I just saw it. I'll add them!

src/Countries/Spain.php
```diff
@@ -6,35 +6,157 @@
...
     }
+
+    /** @return array<string, string> */
+    protected function regionHolidays(int $year): array
+    {
+        if ($year === 2024) {
+            return $this->regionHolidays2024();
+        }
+
+        return [];
```
Nielsvanpach, line 55: I'm not familiar with the holidays in Spain, but isn't it possible to calculate the days instead of hardcoding them? This way we can have correct data for more years.

jaltez:
National, fixed and variable holidays are always the same.

Each autonomous community is free to assign additional holidays that are agreed upon before the beginning of each year. These holidays vary each year, for example, holidays are moved to Monday if they fall on a Sunday, or they are changed to another holiday if there is a long period of time without one.

The problem for these cases is that there are no fixed rules that allow it to be calculated either in the future or in the past, in addition to the fact that each community has different preferences. I think the only thing that can be done is to add some history from recent years, and since we cannot obtain future dates prospectively, add new ones as these are approved.

In my proposal I have tried to group and unify duplicate dates to add some clarity, but to consider regional cases I can't think of many alternatives.

Nielsvanpach:
If it's random every year, adding a bit of historical data seems fine to me.

src/Countries/Spain.php
```diff
@@ -0,0 +1,388 @@
...
+            ] + $april10th + $june24th,
+
+            default => [],
+        };
+    }
+
+    /** @return array<string, string> */
+    protected function regionHolidays2024(): array
+    {
+        $march19th = ['San José' => '03-19'];
```
Nielsvanpach, line 288: Why not name the variables by the holidayName?

tests/Countries/SpainTest.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Spatie\Holidays\Tests\Countries;
+
+use Carbon\CarbonImmutable;
+use Spatie\Holidays\Countries\Spain;
+use Spatie\Holidays\Holidays;
+
+it('can calculate spanish holidays', function () {
+    CarbonImmutable::setTestNowAndTimezone('2024-01-01');
```
Nielsvanpach, line 10:
```suggestion
    CarbonImmutable::setTestNow('2024-01-01');
```

tests/Countries/SpainTest.php
```diff
@@ -0,0 +1,31 @@
...
+
+    expect($holidays)
+        ->toBeArray()
+        ->not()->toBeEmpty();
+
+    expect(formatDates($holidays))->toMatchSnapshot();
+});
+
+it('can calculate spanish regional holidays', function () {
+    CarbonImmutable::setTestNowAndTimezone('2024-01-01');
```
Nielsvanpach, line 22:
```suggestion
    CarbonImmutable::setTestNow('2024-01-01');
```

src/Countries/Spain.php
```diff
@@ -0,0 +1,388 @@
...
+        $easter = $this->easter($year);
+
+        return [
+            'Viernes Santo' => $easter->subDays(2),
+        ];
+    }
+
+    /** @return array<string, string> */
+    protected function regionHolidays(int $year): array
+    {
```
Nielsvanpach, line 50:
```php
$holidays = Holidays::for(Spain::make(year: 2020, region: 'es-an'))->get(); 
```

In this case I think we should throw an exception, as we won't be returning the region based holidays.

Nielsvanpach (commented):
Great, thanks!

src/Countries/Spain.php
jaltez: I see, and implement some direct way to also return national holidays, which are still valid, when there is no regional information for a specific year?

Nielsvanpach: You can use InvalidYear::rang()

Nielsvanpach (approved):
Thanks for your work on this!

PR #37, opened 2024-01-17
WIP: Added state support

marlonbasten:
I started working on adding support for states in countries. This will hopefully fix the issues Germany and India are currently having. ( Issues: #8 #12 )

For now I only added NRW (North-Rhine Westphalia) as a POC. Tests also have to be added.

I would really appreciate the help of the Indian and German communities with feedback, improvements and adding tests and more states. I will add more states myself in the coming hours and tomorrow.

### DIFF
--- a/.phpactor.json
+++ b/.phpactor.json
@@ -1,0 +1,4 @@
+{
+    "$schema": "/Applications/Tinkerwell.app/Contents/Resources/phpactor/phpactor.schema.json",
+    "language_server_phpstan.enabled": false
+}
--- a/src/Countries/Country.php
+++ b/src/Countries/Country.php
@@ -5,6 +5,7 @@ namespace Spatie\Holidays\Countries;
 use Carbon\CarbonImmutable;
 use Spatie\Holidays\Exceptions\InvalidYear;
 use Spatie\Holidays\Exceptions\UnsupportedCountry;
+use Spatie\Holidays\States\State;
 
 abstract class Country
 {
@@ -14,30 +15,35 @@ abstract class Country
     abstract protected function allHolidays(int $year): array;
 
     /** @return array<string, CarbonImmutable> */
-    public function get(int $year): array
+    public function get(int $year, ?State $state = null): array
     {
         $this->ensureYearCanBeCalculated($year);
 
-        $allHolidays = $this->allHolidays($year);
+        $allHolidaysCountry = $this->formatHolidays(
+            holidays: $this->allHolidays($year),
+            year: $year,
+        );
 
-        $allHolidays = array_map(function ($date) use ($year) {
-            if (is_string($date)) {
-                $date = CarbonImmutable::createFromFormat('Y-m-d', "{$year}-{$date}");
-            }
+        $allHolidaysState = $this->formatHolidays(
+            holidays: $state?->allHolidays($year) ?? [],
+            year: $year,
+        );
 
-            return $date;
-        }, $allHolidays);
+        $allHolidays = [
+            ...$allHolidaysCountry,
+            ...$allHolidaysState,
+        ];
 
         uasort($allHolidays,
-            fn (CarbonImmutable $a, CarbonImmutable $b) => $a->timestamp <=> $b->timestamp
+            static fn (CarbonImmutable $a, CarbonImmutable $b) => $a->timestamp <=> $b->timestamp
         );
 
         return $allHolidays;
     }
 
-    public static function make(): static
+    public static function make(?State $state = null): static
     {
-        return new static();
+        return new static($state);
     }
 
     public static function find(string $countryCode): ?Country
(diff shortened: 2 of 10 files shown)


Nielsvanpach:
Hi @marlonbasten, thanks for your effort on this! As you might have noticed, this package is blowing up with PRs, so it takes a while to get through them.

I'm sure your approach works fine for Germany, but we decided to use a simpler solution for states and regions which is extendable for more use cases. I've added some docs on this: https://github.com/spatie/holidays?tab=readme-ov-file#contributing-a-new-country

Let me know what you think.

marlonbasten:
@Nielsvanpach I saw that someone approved the workflows for this PR 😄. I saw that PHPStan was failing so I fixed that. Even if this doesn't get merged, maybe something can be salvaged from this PR.

eusonlito:
How many levels should be added?

In Spain we have national (country), by autonomous community (region), province (state) and city.

Is this a common case in other countries?

marlonbasten:
> How many levels should be added?
> 
> In Spain we have national (country), by autonomous community (region), province (state) and city.
> 
> Is this a common case in other countries?

@eusonlito We have something similar in Germany. But this is only an exception for a few cities. In my solution this could be solved like that:

(Disclaimer: My solution uses the name "state". So I'll call everything a State)

- Create a state class "GaliciaRegion" and add the holidays that apply to that region
- Create a state class "LugoProvince" and make it extend "GaliciaRegion" and merge the holidays of Galicia and Lugo in it
- Create a state class "LugoCity" and make it extend "LugoProvince", do the same as above.

Now you can get all holidays.

Get all national holidays:
```php
Holidays::for(Spain::make())
  ->get();
```

Get holidays from Spain and the Galicia region:
```php
Holidays::for(Spain::make(), state: GaliciaRegion::make())
  ->get();
```

Get holidays from Spain and the Lugo province in the Galicia region:
```php
Holidays::for(Spain::make(), state: LugoProvince::make())
  ->get();
```

Get holidays from Spain and Lugo city in the Lugo province in the Galicia region:
```php
Holidays::for(Spain::make(), state: LugoCity::make())
  ->get();
```

Could this solution work for you?

marlonbasten:
@Nielsvanpach Thanks for running the tests again! I genuinely do now know why this one test is failing. When looking at the logs everything seems to be alright, but the test is flagged as filed.

eusonlito:
@marlonbasten perhaps the solution is for everything to be a region?

And each region can extend to a parent region. This way you can create as many levels as you need depending on the location:

City > State > Region > Country

In any case, you should be able to get only the vacations of a city, without including those of its province, region or country, since in many cases vacations are shown separately in the calendars (local, provincial, region, national).

marlonbasten:
@eusonlito That would be an option too. But what about if you want to get the holidays for a city including the nationwide holidays? In Germany for example we have nationwide (country) holidays and state-specific holidays. 

Those don't get separated and are always shown in one. So maybe I change the following in my PR:
- Change the name "state" to "region"
- Create logic to extend a Region but without merging the holidays of the extended region.

Then you could do the same as before but can now also do this for the regions separately without getting the "parents" holidays.

```php
State::findOrFail(Spain::make(), LugoCity::make())->get();
```

eusonlito:
@marlonbasten yes, by default it is all holidays together, but perhaps an additional method should be added to `make` that returns only holidays specific to that region.

And... how about changing the directory structure for the holidays?

```
src/Locations/Spain/Spain.php // Country
src/Locations/Spain/Andalucia/Andalucia.php // Region
src/Locations/Spain/Andalucia/Sevilla/Sevilla.php // State
src/Locations/Spain/Andalucia/Sevilla/Sevilla/Sevilla.php // City
src/Locations/Spain/Andalucia/Sevilla/Sevilla/DosHermanas.php // City
src/Locations/Spain/Galicia/Galicia.php // Region
src/Locations/Spain/Galicia/Galicia/Lugo/Lugo.php // State
src/Locations/Spain/Galicia/Galicia/Lugo/Lugo/Lugo.php // City
src/Locations/Spain/Galicia/Galicia/Lugo/Lugo/Foz.php // City
```

:grinning:

marlonbasten:
@eusonlito I could do that. But I don't want to put any more time into this PR because I already got told that my solution will not be merged 😅

If any of the core members of Spatie agree to a solution proposed by @eusonlito or me I would be happy to take my time to make those changes! :)

marlonbasten:
@MeiKatz Smart and easy solution indeed! That would work well for Germany. I have an initial thought about that tho:

How would you get only the holidays that apply to that state? As far as I understand, most people in Spain want to show state/federal/etc.
holidays separately. Would you merge that in the state method somehow? Maybe implement a HasStates/HasRegions Interface?

Maybe I'll try to implement that locally and if I like it I'll change that in my PR.

On another note: Are you satisfied with the current official proposal about states/regions?

marlonbasten:
Closing this PR now due to lack of communication.

PR #39, opened 2024-01-17
add support for country regions

Nielsvanpach:


marlonbasten:
Also a good solution! Can you maybe also look at my solution? #37
I think both solutions have certain upsides and downsides. Yours is simple and easier to extend.
Mine is more type-safe when needed and adds the functionality to get only state-specific holidays

PR #43, opened 2024-01-17
Add Australian Holidays

patrickomeara:
These are the national holidays down to the state level.

### DIFF
--- a/src/Countries/Australia.php
+++ b/src/Countries/Australia.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\CarbonImmutable;
+
+class Australia extends Country
+{
+    public function countryCode(): string
+    {
+        return 'au';
+    }
+
+    protected function allHolidays(int $year): array
+    {
+        return array_merge([
+            'New Year\'s Day' => '01-01',
+            'Australia Day' => '01-26',
+            'Anzac Day' => '04-25',
+            'Christmas Day' => '12-25',
+            'Boxing Day' => '12-26',
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        $easter = CarbonImmutable::createFromTimestamp(easter_date($year))->setTimezone('Australia/Sydney');
+
+        return [
+            'Good Friday' => $easter->subDays(2),
+            'Easter Monday' => $easter->addDay(),
+        ];
+    }
+}
--- a/src/Countries/Australia.php
+++ b/src/Countries/Australia.php
@@ -6,6 +6,11 @@ use Carbon\CarbonImmutable;
 
 class Australia extends Country
 {
+    protected function __construct(
+        protected ?string $region = null,
+    ) {
+    }
+
     public function countryCode(): string
     {
         return 'au';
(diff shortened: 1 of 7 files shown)


src/Countries/Australia.php
```diff
@@ -0,0 +1,39 @@
...
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        $easter = CarbonImmutable::createFromTimestamp(easter_date($year))->setTimezone('Australia/Sydney');
+
+        return [
+            'Queen\'s Birthday' => CarbonImmutable::parse("second monday of june $year"),
```
samlev, line 32: It's King's birthday now :)

It's also regional - different dates for different states.

patrickomeara: I've removed it. Once the package handles states we can get a little more granular.

samlev: Yeah, there's something mentioning regional dates in the readme, but regional holidays in Australia get weird

patrickomeara: I've also removed Easter Saturday and Sunday as some states don't observe them.

Let's see what comes of @Nielsvanpach's PR #39.

Nielsvanpach: i've added documentation on how to handle regions: https://github.com/spatie/holidays?tab=readme-ov-file#contributing-a-new-country

src/Countries/Australia.php
```diff
@@ -0,0 +1,35 @@
...
+            'Anzac Day' => '04-25',
+            'Christmas Day' => '12-25',
+            'Boxing Day' => '12-26',
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        $easter = CarbonImmutable::createFromTimestamp(easter_date($year))->setTimezone('Australia/Sydney');
```
Nielsvanpach, line 28: Could you use `$this->easter($year)` instead? Available after rebasing from this repo's main branch.

src/Countries/Australia.php
kurucu: Why not just add all days, and suffix the description with the state in brackets? e.g. "King's Holiday (VIC)".

patrickomeara: I've added regional support to the state level.

src/Countries/Australia.php
patrickomeara: This is done.

Nielsvanpach:
Thanks! I'll leave this open until this has some approvals by others.

samlev (approved):
Looks good to me without going into the absolute absurdity that is "replacement public holidays" (i e. those certain public holidays that fall on a weekend so some states give a "replacement" public holiday on the following Monday)

patrickomeara:
Thanks @samlev I did think about the replacement days but decided against implementing something at this stage. The context of how the package is being used is a big factor to whether replacement days are useful or not.

src/Countries/Australia.php
```diff
@@ -0,0 +1,119 @@
...
+        return [
+            'Good Friday' => $easter->subDays(2),
+            'Easter Monday' => $easter->addDay(),
+            // https://en.wikipedia.org/wiki/Public_holidays_in_Australia
+            ...array_filter(match($this->region) {
+                'act' => [
+                    'Canberra Day' => CarbonImmutable::parse("second monday of march {$year}"),
+                    'Easter Saturday' => $easter->subDay(),
+                    'Easter Sunday' => $easter,
+                    'Reconciliation Day' => CarbonImmutable::create($year, 5, 27)->modify("monday"),
```
patrickomeara, line 44: @Nielsvanpach would you rather this PHPStan failure ignored or a ternary?

```php
'Reconciliation Day' => ($may27 = CarbonImmutable::create($year, 5, 27)) ? $may27->modify("monday") : null,
```

samlev: I would suggest something like:

```php
Carbon::create($year, 5, 27)->modify('monday')->toImmutable()
```

patrickomeara: This gives two errors 😄 

```
  45     Cannot call method modify() on Carbon\Carbon|false.
  45     Cannot call method toImmutable() on Carbon\Carbon|false.
  ```

patrickomeara: Added to the phpstan baseline, there were others in there already so I assume that's the right course of action here.

patrickomeara:
The tests seem a little flaky, when running all the tests locally I get between 1 and 3 fails. Uganda, Estonia and North Macedonia.

Nielsvanpach:
Thanks, PR looks good! I'll have a look at the failing tests.

PR #44, opened 2024-01-17
Add support for Irish public holidays

marchanlon:
Adds support for [Irish public holidays](https://www.citizensinformation.ie/en/employment/employment-rights-and-conditions/leave-and-holidays/public-holidays/) 🇮🇪


Nielsvanpach:
Thanks!

PR #45, opened 2024-01-17
Add support for Croatian public holidays

mcbuckets:
Croatian 🇭🇷 public holiday list: https://neradni-dani.com/national-public-holidays-croatia.php


Nielsvanpach:
Thanks!

PR #49, opened 2024-01-17
Draft: Add Support for Switzerland

Martin-Welte:
## TL;DR

- Adds Support for Switzerland

## Additional Information

until [discussion about region support](https://github.com/spatie/holidays/discussions/14) is wrapped up, this PR includes the holidays for switzerland that the public transport system, the post and nation-wide organisations follow. 

## Source

- [German Wikipedia](https://de.wikipedia.org/wiki/Feiertage_in_der_Schweiz)


src/Countries/Switzerland.php
```diff
@@ -0,0 +1,39 @@
...
+use Carbon\CarbonImmutable;
+
+class Switzerland extends Country
+{
+    public function countryCode(): string
+    {
+        return 'ch';
+    }
+
+    /** @return array<string, string|CarbonImmutable> */
```
Nielsvanpach, line 14: Can be removed

src/Countries/Switzerland.php
```diff
@@ -0,0 +1,39 @@
...
+            'Bundesfeier' => '08-01',
+            'Weihnachtstag' => '12-25',
+            'Stephanstag' => '12-26',
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        $easter = CarbonImmutable::createFromTimestamp(easter_date($year))
```
Nielsvanpach, line 29: Use `$this->easter($year)` instead. Only available if you rebased on this repo's main branch

Nielsvanpach (commented):
Thanks! Could you rebase, that way I can run the tests.

I've also added notes on region based holidays: 

src/Countries/Switzerland.php
Martin-Welte: done

src/Countries/Switzerland.php
Martin-Welte: done

tests/.pest/snapshots/Countries/SwitzerlandTest/it_can_calculate_swiss_holidays.snap
```diff
@@ -0,0 +1,38 @@
+[
+    {
+        "name": "Karfreitag",
+        "date": "1969-12-30"
```
Nielsvanpach, line 4: results don't seem to be correct

src/Countries/Switzerland.php
```diff
@@ -0,0 +1,37 @@
...
+    public function countryCode(): string
+    {
+        return 'ch';
+    }
+
+    protected function allHolidays(int $year): array
+    {
+        return array_merge([
+            'Neujahr' => '01-01',
+            'Berchtoldstag' => '01-02',
```
herpaderpaldent, line 18: This is not a public holiday in every canton. In the resource (wikipedia) you can see that this day is not official everyware. 
F.e. Kt. BE does not celebrate this day.

Martin-Welte: image
@herpaderpaldent this PR covers only the non-regional holidays until now. You can read the description of the PR which days are included and why.

herpaderpaldent: Hey @Martin-Welte i am sorry i can not see a description to this PR that would have explained this. However the simplification is understandable until the regional support is settled. 

May i suggest changing the source towards the original source and not wikipedia?

herpaderpaldent:
Hi @Martin-Welte 

I am sorry to bother you, i already started forking your work and play around with the cantons based on the BFJ List. However something i noticed we haven't discussed: Language.

How is multilanguage support handled with this package? Right now the german variants of the holidays is present but we have 4 official languages in switzerland. 
Any idea regarding this? Maybe @Nielsvanpach ?

Martin-Welte:
@herpaderpaldent there is a discussion here about multi lang support: https://github.com/spatie/holidays/discussions/131

Martin-Welte:
@herpaderpaldent I would love to see your shot at cantons. I am still pondering what the best solution would be here.

herpaderpaldent:
> @herpaderpaldent I would love to see your shot at cantons. I am still pondering what the best solution would be here.

This is what i have been tinkering with tonight. Let me know what you think: https://github.com/herpaderpaldent/holidays/blob/feature/switzerland-with-regions/src/Countries/Switzerland.php

Martin-Welte:
@herpaderpaldent Thank you for your PR. 
Each canton has their own set of holidays, so it's unfortunateley not dependant on a language. I am also working on an approach that could then be translated easily in the future. 
For now i was thinking something along these lines:
```php
$regions = [
  'ZH' => [
    'Neujahr' => '01-01',
    'Karfreitag' => $easter->subDays(2),
    'Ostermontag' => $easter->addDay(),
    'Tag der Arbeit' => '05-01',
    'Auffahrt' => $easter->addDays(39),
    'Pfingstmontag' => $easter->addDays(50),
    'Bundesfeier' => '08-01',
    'Weihnachtstag' => '12-25',
    'Stephanstag' => '12-26',
  ],
];

return $regions[$this->region];
```
that way we could define each canton seperateley, and have it nice and tidy. Instead of using the german words i would opt for translation keys which we can then translate via a seperate locale in the future.

herpaderpaldent:
Hey 

I cannot really follow your comment my attempt allready reflects the fact that every canton has its own set of holidays:

f.e.:
```php
private function getCantonHolidays(string $canton, int $year): array
    {
        return match ($canton) {
            'vs' => $this->getVSHolidays($year),
            default => [],
        };
    }
    
    private function getVSHolidays(int $year)
    {
        $eastern_holidays = $this->getEasterHolidays($year, with_easter_minus_two: false, add_corpus_christi: true);

        return array_merge([
            $this->getLocaleHoliday('01-01') => '01-01',
            $this->getLocaleHoliday('01-02') => '01-02',
            $this->getLocaleHoliday('03-19') => '03-19', // St. Joseph
            $this->getLocaleHoliday('08-15') => '08-15', // Assumption
            $this->getLocaleHoliday('11-01') => '11-01', // All Saints
            $this->getLocaleHoliday('12-08') => '12-08', // Immaculate Conception
            $this->getLocaleHoliday('25-12') => '12-25',
            $this->getLocaleHoliday('26-12') => '12-26',
        ], $eastern_holidays);
    }
```

Although having slept a night i would aim for a better translation approach then the current. F.e. having the Translations in a single constant and use english as key for the translation. Fallback to the first language as default.

```php
const HOLIDAYS = [
        'languages' => ['de', 'fr', 'it', 'rom'],
        'New Year' => ['Neujahrstag', 'Nouvel an', 'Capodanno', 'Nadal'],
.....
];
```

Martin-Welte:
@herpaderpaldent I had time today to update it today, now we can discuss it.
What I wanted to ensure was having the variable holidays clearly defined, so that it's rather obvious which holidays will be used. It did take a bit of typing to ensure it though.

Martin-Welte:
@Nielsvanpach btw, there is a rudimentary translation function in there, if you want to take a look. Right now it only translates into the 3 official languages of switzerland.

herpaderpaldent:
I don't like your solution. Way to much boilerplate. 

Have you even checked out my approach? You have not even given feedback to it

Martin-Welte:
@Nielsvanpach thank you for the info. I will update the PR End of Week

@herpaderpaldent not yet, i had limited time until now, but i will take a look. I wanted to make sure you can see how i pictured my version. I am sorry, but unfortunately I have quite the bussy week. If you have time in the meantime you could comment, what you would do differently and why. Which parts of your implementation would you say, would be most crucial to take into consideration?

Kenny1291:
New Year's Day
Ascension Day
Christmas Day

Kenny1291:
Hello guys, just dropping some info here about public holidays in Switzerland.

Swiss National Day (August 1) is the only federal public holiday. Cantons have their own additional public holidays.
The only holidays that every Canton has in common are: 

- New Year's Day
- Ascension Day
- Christmas Day

[Labour Law](https://www.fedlex.admin.ch/eli/cc/1966/57_57_57/de#lvl_III/lvl_2/lvl_u13)

Nielsvanpach (approved):
Thanks!

PR #56, opened 2024-01-18
Add Moroccan Holidays

YazidKHALDI:
This pull request enhances the support for Moroccan national holidays
**Note:**
- Islamic holidays' dates are subject to change based on the Hijri calendar, and efforts will be made to ensure accurate and up-to-date information in subsequent releases.

Do I have the right to use [geniusts/hijri-dates](https://github.com/GeniusTS/hijri-dates) to convert Date convert Gregorian date to Hijri date and vice versa?
I will try to test it today to check how much it's accurate


src/Countries/Morocco.php
```diff
@@ -0,0 +1,49 @@
...
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        /**
+         * The following holidays are considered public holidays in Morocco. However, their dates vary each year,
+         * as they are based on the Islamic Hijri (lunar) calendar. These holidays do not have a fixed date and
+         * occur based on the lunar calendar sequence. The order listed reflects the chronological occurrence
+         * of these holidays throughout the year.
+         */
+        return [
```
YazidKHALDI, line 39: Do I have the right to use `geniusts/hijri-dates` to convert Date   convert Gregorian date to Hijri date and vice versa ? 
i will try to test it today  to check how much it's accurate

medilies: It is a necessary feature to not exclude half the holidays from Muslim countries

Nielsvanpach: I am not really familiar with this. Could you shortly explain what this package would do and how we could benefit from it?

YazidKHALDI: The [geniusts/hijri-dates](https://github.com/GeniusTS/hijri-dates) package is a reliable PHP library designed to facilitate the conversion between Gregorian and Hijri dates. In our case, it can be incredibly useful for handling Islamic holidays that are based on the Hijri calendar.

The Islamic calendar follows a lunar system, which means that its months and years are determined by the moon's phases. This leads to variations in the lengths of months and the occurrence of Islamic holidays. By integrating this package, we gain the ability to accurately convert dates between the Gregorian and Hijri calendars.

I am currently in the process of testing the package to ensure its efficacy. I have a slight doubt about its accuracy, but my friend from the Maldives has used it in his [PR#63](https://github.com/spatie/holidays/pull/63/). I will conduct thorough testing today to evaluate its performance and accuracy.

YazidKHALDI: Hi @Nielsvanpach,

I utilized the [geniusts/hijri-dates](https://github.com/GeniusTS/hijri-dates) package to compute Islamic holidays for Morocco. I have included tests for the years 2023, 2024, 2025, and 2026. The results match the date converter provided by the [Minister of Religious Endowments and Islamic Affairs](https://www.habous.gov.ma/%D9%85%D8%AD%D9%88%D9%84-%D8%A7%D9%84%D8%AA%D8%A7%D8%B1%D9%8A%D8%AE).

Please note that it's not possible to predict Islamic holidays with 100% certainty. Ministry observers use the naked-eye observation of the [New Moon](https://en.wikipedia.org/wiki/New_Moon) at the end of each lunar month to determine if the current month is 29 or 30 days, impacting the declaration of the new month.

Nielsvanpach: Thanks for that! Just for clarity; do countries with Islamic public holidays all have the same dates for them? So is `Eid al-Fitr` celebrated everywhere on the same day, or are there exceptions?

If so, is it possible to not limit this to only Morocco, but make the calculations accessible for all countries with Islamic holidays?

medilies: @Nielsvanpach It is common for Islamic countries but the phase of the moon is not the same everywhere; For example `Eid Al-Fitr` may occur on the 10th of April for Algeria and on the 11th of April for Morocco.

Btw, this #90 is also making use of the same package.

YazidKHALDI: @Nielsvanpach 

The topic is more intricate than it seems, and there are ongoing debates in Islamic countries. There are two main differences between countries:

1. **Observation Methods:** Not all countries share the same date due to differences of approximately ±1 day. This is because each country employs its own method to observe the new moon, whether through naked-eye observation, telescopes, or astronomical calculations. Some countries find it impossible to determine the new moon using the naked eye.

   As an illustration, consider the current Hijri month and the next one. If you check the Hijri date today in all Islamic countries, it will be 09/07/1445. To understand why they all have the same date, check this image, representing the visibility of the moon on the first day of the current Hijri month "Rajab" ([source](https://www.moonsighting.com/1445rjb.html)). 
   
   In this image, all countries easily observe the new moon with the naked eye, leading them to declare the 13th of January 2024 as the first day of the month "Rajab."


   Now, observe the visibility of the moon on the first day of the next Hijri month in this image ([source](https://www.moonsighting.com/1445shb.html)).
   
   You can see that only the countries of West Africa (Morocco, Algeria, etc.) can easily observe the new moon. They declare the 11th of February 2024 as the first day of the month "Shaban." However, Egypt and other Middle Eastern countries may face challenges in seeing the new moon. Consequently, some will declare February 11, 2024, as the first day based on astronomical calculations, while others will wait until February 12, 2024.

2. **Holiday Duration:** Another difference is the number of holidays and the number of days taken as holidays. For example, "Eid al-Fitr" is observed for just 2 days in Morocco, but it's 3 days in Algeria and 4 days in Saudi Arabia.

In conclusion, having the same date for all countries is challenging, and achieving 100% accuracy in calculations is difficult because each country has its own protocol for the observation of the new moon.

PS: I fixed the PHPStan failure.

ahmedmansoor:
I've created a new PR #186, (originally #63), in which I've added a Hijri Calendar trait that we can use [geniusts/hijri-dates](https://github.com/GeniusTS/hijri-dates) to calculate Hijri dates.

YazidKHALDI:
@Nielsvanpach 
HI Niels
I hope you're doing well. I've made some updates to the project,  , I've removed the dependency on the geniusts/hijri-dates package and integrated the calculation function provided by the [Ministry of Endowments and Islamic Affairs in Morocco](https://www.habous.gov.ma/%D9%85%D8%AD%D9%88%D9%84-%D8%A7%D9%84%D8%AA%D8%A7%D8%B1%D9%8A%D8%AE)  .

This change simplifies the codebase and reduces external dependencies, 

thank you

src/Countries/Morocco.php
```diff
@@ -0,0 +1,153 @@
...
+    use Translatable;
+
+    public function countryCode(): string
+    {
+        return 'ma';
+    }
+
+    public function defaultLocale(): string
+    {
+        return 'en';
```
Nielsvanpach, line 21: shouldn't this be Arabic by default?

src/Countries/Morocco.php
```diff
@@ -0,0 +1,153 @@
...
+            'Youth Day' => '08-21',
+            'Green March' => '11-06',
+            'Independence Day' => '11-18',
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        // Calculate the current Hijri year based on the Gregorian year
```
Nielsvanpach, line 43: You can have a look at other Islamic countries and how their lunar based holidays are resolved. For example: Turkey, Bahrein and Egypt.

Nielsvanpach (changes requested):
Thanks for your efforts on this!

src/Countries/Morocco.php
YazidKHALDI: thanx for the review 

In fact, I assumed that the Package would be used globally and therefore set English as the default language.

src/Countries/Morocco.php
YazidKHALDI: I've reviewed the other contributions, and it seems they've hardcoded the expected dates of Islamic holidays. Personally, I don't find that approach ideal. Instead, I opted to use a date converter for more dynamism. While it may not be 100% accurate, it offers flexibility. If you still prefer the hardcoded holiday expectation dates, I'm willing to make the change tonight.

Nielsvanpach (approved):
Thanks!

PR #58, opened 2024-01-18
Add English Holidays

ashnewmanjones:
This adds the English & Welsh holidays according to the Gov website: https://www.gov.uk/bank-holidays

### DIFF
--- a/src/Countries/England.php
+++ b/src/Countries/England.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\CarbonImmutable;
+
+class England extends Country
+{
+  public function countryCode(): string
+  {
+    return 'en';
+  }
+
+  protected function allHolidays(int $year): array
+  {
+    return array_merge([
+      'New Year\'s Day' => new CarbonImmutable($year . "-01-01", 'Europe/London'),
+      'Early May bank holiday' => new CarbonImmutable("first monday of may {$year}", 'Europe/London'),
+      'Spring bank holiday' => new CarbonImmutable("last monday of may {$year}", 'Europe/London'),
+      'Summer bank holiday' => new CarbonImmutable("last monday of august {$year}", 'Europe/London'),
+      'Christmas Day' => new CarbonImmutable($year . "-12-25", 'Europe/London'),
+      'Boxing Day' => new CarbonImmutable($year . "-12-26", 'Europe/London'),
+    ], $this->variableHolidays($year));
+  }
+
+  protected function variableHolidays(int $year): array
+  {
+    $easterSunday = CarbonImmutable::createFromTimestamp(easter_date($year))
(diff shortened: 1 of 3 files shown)


tests/Countries/EnglandTest.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Spati\Holidays\Countries;
```
nexxai, line 3: I believe the `Spatie` is missing an `e` here

SimonBarrettACT:
This doesn't account a bank holiday falling on a weekend, a ‘substitute’ weekday becomes a bank holiday, normally the following Monday.

ashnewmanjones:
That is a good point, given that the Welsh and English holidays are the same it's probably worth just using your PR #75. Not sure on the best way to support en and cy without duplicating all the code though.

SimonBarrettACT:
> That is a good point, given that the Welsh and English holidays are the same it's probably worth just using your PR #75. Not sure on the best way to support en and cy without duplicating all the code though.

I've added English holidays to PR #75.

```
<?php

namespace Spatie\Holidays\Countries;

class England extends Wales
{
    public function countryCode(): string
    {
        return 'en';
    }
}

```

src/Countries/England.php
```diff
@@ -0,0 +1,39 @@
...
+  public function countryCode(): string
+  {
+    return 'en';
+  }
+
+  protected function allHolidays(int $year): array
+  {
+    return array_merge([
+      'New Year\'s Day' => new CarbonImmutable($year . "-01-01", 'Europe/London'),
+      'Early May bank holiday' => new CarbonImmutable("first monday of may {$year}", 'Europe/London'),
```
Nielsvanpach, line 18: Could you use the format used in the other Country classes? Also, this seems to be a variable holiday. Makes more sense to put in the dedicated function? Same for Spring bank & Summer bank, I think

src/Countries/England.php
```diff
@@ -0,0 +1,39 @@
...
+      'Spring bank holiday' => new CarbonImmutable("last monday of may {$year}", 'Europe/London'),
+      'Summer bank holiday' => new CarbonImmutable("last monday of august {$year}", 'Europe/London'),
+      'Christmas Day' => new CarbonImmutable($year . "-12-25", 'Europe/London'),
+      'Boxing Day' => new CarbonImmutable($year . "-12-26", 'Europe/London'),
+    ], $this->variableHolidays($year));
+  }
+
+  protected function variableHolidays(int $year): array
+  {
+    $easterSunday = CarbonImmutable::createFromTimestamp(easter_date($year))
```
Nielsvanpach, line 28: Could you use `$this->easter($year)` instead? Available after having your branch rebased on our repo's main branch.

Nielsvanpach (commented):
Thanks!

ashnewmanjones:
Use PR #75 as is more complete

PR #74, opened 2024-01-18
Add holidays for Finland

ConnySjoblom:
This PR adds Finnish holidays!

Would also be nice to see support for languages here. So having `languageCode()` on top of `countryCode()` for instance. Finland for instace has both Swedish (se_FI) and Finnish (fi_FI) as official languages.


src/Countries/Finland.php
```diff
@@ -0,0 +1,54 @@
...
+use Carbon\CarbonImmutable;
+
+class Finland extends Country
+{
+    public function countryCode(): string
+    {
+        return 'fi';
+    }
+
+    /** @return array<string, CarbonImmutable> */
```
Nielsvanpach, line 14: you can remove this

Nielsvanpach (commented):
Thanks! Could you rebase from main, so I can run the tests here?

Nielsvanpach:
> Would also be nice to see support for languages here. So having languageCode() on top of countryCode() for instance. Finland for instace has both Swedish (se_FI) and Finnish (fi_FI) as official languages.

I'm open to a PR which adds language support. Would be best in a different PR.

src/Countries/Finland.php
ConnySjoblom: Done

ConnySjoblom:
git settings apparently wrong on second machine, will reopen.

PR #76, opened 2024-01-18
Class and test for croatian (hr) holidays

erikroznbeker:
Holidays source is from site https://neradni-dani.com/kalendar-2024.php

### DIFF
--- a/src/Countries/Croatia.php
+++ b/src/Countries/Croatia.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\CarbonImmutable;
+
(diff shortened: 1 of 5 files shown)


src/Countries/Croatia.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\CarbonImmutable;
+
+class Croatia extends Country
+{
+    protected function __construct(
```
Nielsvanpach, line 9: No need for the constructor, when not used

Nielsvanpach:
Thanks, but closing this one, as a Croatia PR has already been merged.

PR #77, opened 2024-01-18
Add Bangladeshi holidays

iRaziul:


Nielsvanpach:
Hi there, we already merged a Pr for Bangladesh. It seems like you have some differences with our current implementation. Could you verify which one is correct?

PR #82, opened 2024-01-18
[Fix] Enabling `Brazil` Holidays

devajmeireles:


devajmeireles:
@Nielsvanpach,

There is something very strange. Locally all tests pass, and CI fails.

Locally I've tested with all PHP versions:

8.1:
CleanShot 2024-01-18 at 11 16 17

8.2:
CleanShot 2024-01-18 at 11 16 34

8.3:
CleanShot 2024-01-18 at 11 16 49

devajmeireles:
I'm wondering if we don't need `--update-snapshots` on CI? 🤔

Nielsvanpach:
Not really. The results in the snapshot file should be verified before merging & remain unchanged after merging.

Nielsvanpach:
Thanks. We fixed this in the meantime!

PR #89, opened 2024-01-18
Add Angola Holidays 

monteirofutila:
🇦🇴

### DIFF
--- a/src/Countries/Angola.php
+++ b/src/Countries/Angola.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\CarbonImmutable;
+
+class Angola extends Country
+{
+    public function countryCode(): string
+    {
+        return 'ao';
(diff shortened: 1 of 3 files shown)


src/Countries/Angola.php
```diff
@@ -0,0 +1,42 @@
...
+    public function countryCode(): string
+    {
+        return 'ao';
+    }
+
+    protected function allHolidays(int $year): array
+    {
+
+        return array_merge([
+            'Dia de Ano Novo' => '01-01',
```
Nielsvanpach, line 18: TIL, Angola speaks Portuguese!

Nielsvanpach (commented):
PR looks good, thanks! I'll wait for some confirmation of others before merging.

src/Countries/Angola.php
monteirofutila: Yes, Portuguese is the official language of Angola.

PR #90, opened 2024-01-18
Add Algerian holidays

celyes:
Added official holidays for Algeria.


composer.json
```diff
@@ -21,8 +21,9 @@
     ],
     "require": {
         "php": "^8.1",
-        "nesbot/carbon": "^2.72.1",
-        "ext-calendar": "*"
+        "ext-calendar": "*",
+        "geniusts/hijri-dates": "^1.1",
```
Nielsvanpach, line 25: Not using this library for a couple of reasons:
1. https://github.com/spatie/holidays/pull/155#issuecomment-1904467312
2. https://github.com/spatie/holidays/blob/main/src/Countries/Turkey.php#L11

src/Countries/Algeria.php
```diff
@@ -0,0 +1,123 @@
...
+    protected function allHolidays(int $year): array
+    {
+        return array_merge($this->fixedHolidays($year), $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function fixedHolidays(int $year): array
+    {
+        return [
+            'New Year\'s Day' => CarbonImmutable::createFromDate($year, 1, 1),
```
Nielsvanpach, line 26: Can you use the format like in other classes? 
Look at Belgium.php for example

Nielsvanpach (changes requested):

Nielsvanpach:
Thanks for your efforts, but as stated before we're not using that package.
Feel free to reopen if you're still planning to update this PR!

PR #91, opened 2024-01-18
Add Japanese Holidays

wadakatu:
Hope it helps.


tests/.pest/snapshots/Countries/JapanTest/it_can_calculate_japanese_holidays.snap
```diff
@@ -0,0 +1,66 @@
+[
+    {
+        "name": "\u5143\u65e6",
```
wadakatu, line 3: I have encountered issues with garbled text due to the use of multibyte characters for holiday names. Should I switch the holiday names from Japanese to English to avoid this problem?

Nielsvanpach:
I've already merged a PR for Japan. Could you verify if the current implementation is correct? Feel free top open a new PR with fixes! Thanks!

PR #92, opened 2024-01-18
Adding Nicaragua Holidays

calonzolg:
Avoid PHPStorm insert final newline into snap, failing the test match string.


freekmurze:
Thanks!

PR #95, opened 2024-01-18
Add Moldavian Holidays

vkosachev:
These are Moldavian Holidays, based on the official dates set by the government
[Reference](https://publicholidays.eu/moldova/ro/2024-dates/)

Since there was some misunderstanding , the Eastern Orthodox Easter is calculated using 
[Meeus' Julian Algorithm](https://en.wikipedia.org/wiki/Date_of_Easter#Meeus's_Julian_algorithm) 
[php_realisation](https://www.php.net/manual/en/function.easter-date.php#83794)


src/Countries/Moldova.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $a = $year % 4;
+        $b = $year % 7;
+        $c = $year % 19;
+        $d = (19 * $c + 15) % 30;
+        $e = (2 * $a + 4 * $b - $d + 34) % 7;
+        $month = intval(floor(($d + $e + 114) / 31));
+        $day = (($d + $e + 114) % 31) + 1;
+
+        $gregorianEaster = CarbonImmutable::createFromDate($year, $month, $day);
```
herpaderpaldent, line 51: First the orthodox calender actually uses the julian calender not the greogrian hence your variable is named wrongly. However if you want to get orthodox eastern you can use php's build in predefined constants:

```php
return CarbonImmutable::createFromTimestamp(easter_date($year, CAL_EASTER_ALWAYS_JULIAN))
```

herpaderpaldent: you are correct i am sorry about the confusion

Nielsvanpach:
There is already a `orthodoxEaster()` function in Country.php
Do you consider that to be a good calculation? If not, why exactly should your calculation be used ad could you provide some tests to verify?

vkosachev:
> There already is a `orthodoxEaster()` function in Country.php Do you consider that to be a good calculation? If not, why exactly should your calculation be used ad could you provide some tests to verify?


@Nielsvanpach, you're correct. The feature wasn't present when I last updated. It appears to have been included in the `a69a226e` commit. The base method now performs calculations in a more brief and concise manner. Thanks for pointing that out.

tests/.pest/snapshots/Countries/MalawiTest/it_can_calculate_ugandan_holidays.snap
```diff
@@ -1,46 +0,0 @@
-[
```
Nielsvanpach, line 1: Why is this removed?

Nielsvanpach (commented):
PR looks good, thanks! 
I'll leave it open until I see some approvals by others on the generated dates.

tests/.pest/snapshots/Countries/MalawiTest/it_can_calculate_ugandan_holidays.snap
vkosachev: That looks like it has been in the wrong directory (Ugandan holidays in Malawi) for some reason. In the snapshots dir Malawi and Ugandan holidays snaps are present.  Just checked that : 
image

Nielsvanpach: Correct! I already fixed it on the main branch too 👍

Nielsvanpach:
In case you need multi language support:
https://github.com/spatie/holidays?tab=readme-ov-file#getting-holidays-in-a-specific-language

vkosachev:
> In case you need multi language support: https://github.com/spatie/holidays?tab=readme-ov-file#getting-holidays-in-a-specific-language

Thanks! @Nielsvanpach  what is the intention around this ? Is this to support official only languages (defined by law) or just to add support for local minorities?

Nielsvanpach:
Both are possible.

Nielsvanpach:
Thanks! Feel free to still send a PR with translations.

PR #109, opened 2024-01-18
add Tunisia Holidays

mr-wolf-gb:
i think we need to use other packages to calculate Hijri holidays like this package [hijri-dates](https://github.com/GeniusTS/hijri-dates), when this package added, i will add the Hijri holidays


src/Countries/Tunisia.php
```diff
@@ -0,0 +1,2266 @@
...
+use Carbon\CarbonImmutable;
+use RuntimeException;
+
+class Tunisia extends Country
+{
+    /**
+     * @var array<int, array<int, array{name: string,date: string,slug: string}>> $hijriHolidays
+     */
+    protected array $hijriHolidays = [
+        "1970" => [
```
Nielsvanpach, line 14: isn't a solution like Turkey possible here?

gaiththewolf: It's not a good solution I know I found the turkey code approved by the repo team so I adapted the code with the other Hijri holidays

gaiththewolf:
I think the team of repo want to use the minimum of required packages

src/Countries/Tunisia.php
gaiththewolf: We can use an external hijri converter package

gaiththewolf: > isn't a solution like Turkey possible here?
> 
> https://github.com/spatie/holidays/blob/main/src/Countries/Turkey.php#L11

Tomorrow I will apply the same Turkey code with more holidays

mr-wolf-gb:
@Nielsvanpach i think these changes not viewed 😅

src/Calendars/HijriCalendar.php
```diff
@@ -0,0 +1,375 @@
+<?php
+
+namespace Spatie\Holidays\Calendars;
+
+use Carbon\CarbonImmutable;
+use RuntimeException;
+
+trait HijriCalendar
```
Nielsvanpach, line 8: As I understand correctly, these dates are not the same for all Islamic countries?

In that case, this class can probably not be reused, so I think it's better to move them to the Tunisia class.

src/Countries/Tunisia.php
```diff
@@ -0,0 +1,82 @@
...
+use Carbon\CarbonImmutable;
+use Spatie\Holidays\Calendars\HijriCalendar;
+
+class Tunisia extends Country
+{
+    use HijriCalendar;
+
+    /**
+     * @return string
+     */
```
Nielsvanpach, line 14: No need for these docblocks when the type is already specified. Only array structures should be typed.

src/Countries/Tunisia.php
```diff
@@ -0,0 +1,82 @@
...
+    public function countryCode(): string
+    {
+        return 'tn';
+    }
+
+    /**
+     * @param int $year
+     * @return array|CarbonImmutable[]|string[]
+     */
```
Nielsvanpach, line 23:
```suggestion
```

src/Countries/Tunisia.php
```diff
@@ -0,0 +1,82 @@
...
+    protected function allHolidays(int $year): array
+    {
+        $revolutionHoliday = [];
+
+        if (2022 > $year and $year >= 2011) {
```
Nielsvanpach, line 28:
```suggestion
        if (2022 > $year && $year >= 2011) {
```

tests/Countries/TunisiaTest.php
```diff
@@ -0,0 +1,45 @@
...
+
+    expect($holidays)
+        ->toBeArray()
+        ->not()->toBeEmpty();
+
+    expect(formatDates($holidays))->toMatchSnapshot();
+});
+
+it('can calculate tunisian holidays 2025', function () {
+    CarbonImmutable::setTestNowAndTimezone('2025-01-01');
```
Nielsvanpach, line 21:
```suggestion
    CarbonImmutable::setTestNow('2025-01-01');
```

tests/Countries/TunisiaTest.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Spatie\Holidays\Tests\Countries;
+
+use Carbon\CarbonImmutable;
+use Spatie\Holidays\Holidays;
+
+it('can calculate tunisian holidays 2024', function () {
+    CarbonImmutable::setTestNowAndTimezone('2024-01-01');
```
Nielsvanpach, line 9:
```suggestion
    CarbonImmutable::setTestNow('2024-01-01');
```

Nielsvanpach (changes requested):

src/Calendars/HijriCalendar.php
mr-wolf-gb: the difference generally in one day, but i changed it to Tunisia only and i removed the trait

Nielsvanpach:
Can you have a look at the failing phpstan errors?

mr-wolf-gb:
> Can you have a look at the failing phpstan errors?

Done

mr-wolf-gb:
@Nielsvanpach any suggestions here !

tests/Countries/TunisiaTest.php
mr-wolf-gb: I think everything is OK

tests/.pest/snapshots/Countries/TunisiaTest/it_can_calculate_tunisia_holidays_with_data_set___2024____2024_.snap
```diff
@@ -12,31 +12,19 @@
         "date": "2024-04-09"
     },
     {
-        "name": "Eid al-Fitr 1. Day",
-        "date": "2024-04-10"
-    },
-    {
-        "name": "Eid al-Fitr 2. Day",
-        "date": "2024-04-11"
+        "name": "Eid al-Fitr",
```
Nielsvanpach, line 15: Have you verified the updated snapshots? Are they correct?

Nielsvanpach (commented):
Looks good!

tests/.pest/snapshots/Countries/TunisiaTest/it_can_calculate_tunisia_holidays_with_data_set___2024____2024_.snap
mr-wolf-gb: i fixed it, it's translation problem

PR #114, opened 2024-01-19
Add Greece holidays

lowv-developer:


src/Countries/Greece.php
```diff
@@ -0,0 +1,81 @@
...
+
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+
+        $orthodox_easter = CarbonImmutable::createFromTimestamp(
+            $this->calculateOrthodoxEaster($year)
```
Nielsvanpach, line 35: You can use `orthodoxEaster()` in Country.php instead

lowv-developer: @Nielsvanpach The orthodoxEaster() is not calculating correct. Im getting wrong date of Easter.

not sure if this is working correct: easter_date($year, CAL_EASTER_ALWAYS_JULIAN);

In https://www.php.net/manual/en/function.easter-date.php gives the correct function to calculate the correct date + small fix. the same function I used in calculateOrthodoxEaster() 

Also Im getting error on other eastern orthodox countries. Orthodox Eastern for 2024 is in 5/5/2024 and the next day is holiday too 5/6/2024

lowv-developer: this is the correct way of calculate orthodox Easter

protected function calculateOrthodoxEaster(int $year): int
    {
      $a = $year % 4;
      $b = $year % 7;
      $c = $year % 19;
      $d = (19 * $c + 15) % 30;
      $e = (2 * $a + 4 * $b - $d + 34) % 7;
      $month = (int) (($d + $e + 114) / 31);
      $day = (($d + $e + 114) % 31) + 1;
      // julian to gregorian
      $jtg = (int) ($year / 100) - (int) ($year / 400) - 2;
      $easterDate = mktime(0, 0, 0, $month, $day + $jtg, $year);
      return (int) $easterDate;
    } // returns timestamp

src/Countries/Greece.php
```diff
@@ -0,0 +1,58 @@
...
+            'Σύναξη της Θεοτόκου' => '12-26',
+
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+
+        $orthodox_easter = $this->orthodoxEaster($year);
```
Nielsvanpach, line 33: Could you use camelCase for PHP variable names?

Nielsvanpach (commented):
Thanks! I'll merge this if the results in snapshot have been verified by someone else too.

src/Countries/Greece.php
lowv-developer: camelCase done! 

I think @dyanakiev from Bulgarian PR #132 can confirm that the current easternOrthodox() is not working

lowv-developer: My statement that it is not working it is not correct. 
I managed to reproduce correct easter date by adding ->setTimezone("Europe/Athens"); in the country.php
it is working correctly only if you add timezone in the $orthodoxEaster = $this->orthodoxEaster($year)->setTimezone("Europe/Athens");

lowv-developer: I notice that NorthMacedonia was passing the tests but for some reason they have timezone +1.

I will change this PR again. sorry for the mess.

lowv-developer: We are using orthodoxEaster in Greece. It is calculated differently. 

In my last commit seems to be working fine. the only downside is that orthodox easter needs setTimezone to work properly.

Yesterday I tried a diff approach but I changed the orthodoxEaster function. I took this back but I added setTimezone. 
Pls check my last commit

lowv-developer: Using startOfDay() on orthodoxEaster will result -1 day on some cases

lowv-developer: Try this in tinker: [https://gist.github.com/lowv-developer/5005f38930002e0a2620cfeb0984665e](url)

lowv-developer: I dont know whats going on. I see other eastern orthodox countries to pass all the tests. There is some kind of bug there but I cannot reproduce exactly. it has something to do with the time and timezones. I'm closing this.

PR #126, opened 2024-01-19
Added Uganda Public Holidays and  Unit Tests

Crawford30:
References: 
1. https://www.edarabia.com/uganda/public-holidays
2. https://africapay.org/uganda/labour-laws/work-time-holiday/holidays-and-work


freekmurze:
Thanks!

PR #132, opened 2024-01-19
Support for Bulgarian holidays

dyanakiev:


src/Countries/Bulgaria.php
```diff
@@ -0,0 +1,82 @@
...
+        ];
+    }
+
+    /**
+     * Calculate the date of Easter using the Computus algorithm with Carbon
+     * @param int $year
+     * @return CarbonImmutable
+     * @throws \RuntimeException if the calculated date is invalid
+     */
+    protected function calculateEasterDate($year): CarbonImmutable
```
Nielsvanpach, line 55: Is this an orthodox calculation? We already have one in Country.php. Does that one work too?

dyanakiev: I totally didn't see that. Yes it works for Bulgaria too.

Nielsvanpach: I think you used the wrong one.

dyanakiev: Yep you're right.. Its correct for 2024 `5 May 2024`, but for 2025 it should be `20 April 2025`. I'm not sure which calculation to use..

dyanakiev: Ok using the `easter_date` function directly actually resolves the issue.

dyanakiev: Ok, nevermind.. 🤣  It's now wrong for 2024, im confused..

Nielsvanpach: Country.php has a easter() and orthodoxEaster() function. I suggest to use one of those. 

If the orthodox calculation gives a false result, feel free to update it, in order to make it work. Do verify if the results for NorthMacedonia is still correct.

dyanakiev: Neither functions give proper result, i will investigate more.

lowv-developer: The problem is that this function in Country.php is wrong calculating the OrthodoxEastern. Please check Greece #114 to fix the problem. @Nielsvanpach needs to update orthodoxEaster() 

> Country.php has a easter() and orthodoxEaster() function.

Nielsvanpach: It's still the wrong function you're calling. You should use orthodoxEaster()

src/Countries/Bulgaria.php
```diff
@@ -0,0 +1,49 @@
...
+            'Съединението на България' => '09-06',
+            'Ден на независимостта на България' => '09-22',
+            'Коледа ден 1' => '12-24',
+            'Коледа ден 2' => '12-25',
+            'Коледа ден 3' => '12-26',
+        ], $this->variableHolidays($year));
+    }
+
+    /**
+     * @param int $year
```
Nielsvanpach, line 36:
```suggestion
```

src/Countries/Bulgaria.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    public function countryCode(): string
+    {
+        return 'bg';
+    }
+
+    /**
+     * @param int $year
+     * @return array<string, string>
+     */
```
Nielsvanpach, line 17:
```suggestion
```

src/Countries/Bulgaria.php
```diff
@@ -0,0 +1,49 @@
...
+    protected function variableHolidays(int $year): array
+    {
+        $easter = CarbonImmutable::createFromTimestamp(easter_date($year))
+            ->setTimezone('Europe/Sofia');
```
Nielsvanpach, line 42:
```suggestion
        $easter = $this->orthodoxEaster($year);
```

tests/.pest/snapshots/Countries/BulgariaTest/it_can_calculate_bulgarian_holidays.snap
```diff
@@ -0,0 +1,46 @@
+[
+    {
+        "name": "\u041d\u043e\u0432\u0430 \u0433\u043e\u0434\u0438\u043d\u0430",
```
Nielsvanpach, line 3: If needed, you ca also add a translation file to the `lang` directory

Nielsvanpach (changes requested):

Nielsvanpach:
Could you have a look at the failing test?

PR #136, opened 2024-01-19
Added Sri Lankan holidays

AvishkaSen:
Disclaimer: 

- There are some full moon (Poya) holidays that can fall a day before or after the actual Gregorian calendar full moon. It's decided by a committee in Sri Lanka. But most of the time it falls on the actual full moon day. 

I will do some research into making it a bit more accurate.

You can check more into it here: https://en.wikipedia.org/wiki/Poya#Poya


AvishkaSen:
Sorry for closing the PR and then reopening it. 
Misclicked the button.

Aside that I was working towards adding a Lunar Full Moon Holiday calculator to the code, since in Sri Lanka its about 12-13 days of holidays based on the full moon. 

The issue is, my code works for the Gregorian days of the Full Moon.
However in Sri Lanka the holidays have a tendency to go a day forward or a day behind the actual full moon date. This happens like 20% of the time. 

Turns out, there's a committee in SL that decides that and does their own calculation based on a different type of calendar. And then calls the full moon holidays a Poya day. 

I contacted some government departments to see what I can learn. 
Until then, I removed the calculation, and some other variable holidays based on different calendars and 3rd party decisions. 

I left in the ones that won't change :)

PR #55, opened 2024-01-18
Added Colombia Holidays

alvleont:
Added the Colombia's holidays. There's a situation with some holidays due to the "Emiliany Law" they are moved to the following monday if it is not monday, so a function to deal with that was added.

PR #141, opened 2024-01-20
Add Colombian holidays

luisprmat:

Yes, i have checked and I know [PR#55](https://github.com/spatie/holidays/pull/55) exists for the same country, but it has problems with `phpstan`. I have rewritten the function that generated problems with types and now the `phpstan` analysis passes.

[PR#55](https://github.com/spatie/holidays/pull/55) can be *closed* in favour of this.

### DIFF
--- a/src/Countries/Colombia.php
+++ b/src/Countries/Colombia.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\CarbonImmutable;
+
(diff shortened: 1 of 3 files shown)


Massoud114:
Done. https://github.com/spatie/holidays/pull/145

Nielsvanpach:
Could you help https://github.com/spatie/holidays/pull/55 with fixing it instead? Thanks!

PR #144, opened 2024-01-20
Update Holidays.php

OnyileFelix1:

### DIFF
--- a/src/Holidays.php
+++ b/src/Holidays.php
@@ -8,15 +8,12 @@ use Spatie\Holidays\Countries\Country;
 
 class Holidays
 {
-    /** @var array<string, CarbonImmutable> */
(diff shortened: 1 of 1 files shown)


OnyileFelix1:
just wanted to add a repository to it

src/Holidays.php
```diff
@@ -84,14 +79,19 @@ public function getName(CarbonInterface|string $date, Country|string|null $count
         return null;
     }
 
+    public function addCustomHoliday(string $name, string $date): self
+    {
+        $this->holidays[$name] = CarbonImmutable::parse($date);
+
+        return $this;
+    }
+
```
OnyileFelix1, line 88: /** @return array<array{name: string, date: CarbonImmutable}> */

PR #145, opened 2024-01-20
Add Benin Holidays

Massoud114:

* [x] Yes, I've checked, and there is no current PR for Benin. I had already opened one and it was closed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,8 +22,7 @@
     "require": {
         "php": "^8.1",
         "nesbot/carbon": "^2.72.1",
-        "ext-calendar": "*",
-        "geniusts/hijri-dates": "^1.1"
+        "ext-calendar": "*"
     },
     "require-dev": {
         "laravel/prompts": "^0.1.15",
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,8 @@
     "require": {
         "php": "^8.1",
         "nesbot/carbon": "^2.72.1",
-        "ext-calendar": "*"
+        "ext-calendar": "*",
+        "geniusts/hijri-dates": "^1.1"
     },
     "require-dev": {
         "laravel/prompts": "^0.1.15",
--- a/src/Countries/Benin.php
+++ b/src/Countries/Benin.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\Carbon;
+use Carbon\CarbonImmutable;
+use GeniusTS\HijriDate\Date;
+use GeniusTS\HijriDate\Hijri;
+
+class Benin extends Country
+{
+    public function countryCode(): string
+    {
+        return 'bj';
+    }
+
+    protected function allHolidays(int $year): array
+    {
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -22,7 +22,8 @@
     "require": {
         "php": "^8.1",
         "nesbot/carbon": "^2.72.1",
-        "ext-calendar": "*"
+        "ext-calendar": "*",
+        "geniusts/hijri-dates": "^1.1"
```
Nielsvanpach, line 26: I prefer to not use this package:
https://github.com/spatie/holidays/pull/155#issuecomment-1904467312

You can have a look at Turkey and how they solved Islamic holidays:

src/Countries/Benin.php
```diff
@@ -0,0 +1,50 @@
...
+            'Jour de Noël' => '12-25',
+            'Jour de l\'assomption' => '08-15',
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        $easter = CarbonImmutable::createFromTimestamp(easter_date($year))
+            ->setTimezone('Africa/Porto-Novo');
```
Nielsvanpach, line 34:
```suggestion
        $easter = $this->easter($year);
```

tests/.pest/snapshots/Countries/BeninTest/it_can_calculate_benin_holidays.snap
```diff
@@ -0,0 +1,54 @@
+[
+    {
+        "name": "F\u00eate du Ramadan",
```
Nielsvanpach, line 3: Results seem to be wrong? Have you checked all results in the snapshot?

tests/Countries/BeninTest.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Spatie\Holidays\Tests\Countries;
+
+use Carbon\CarbonImmutable;
+use Spatie\Holidays\Holidays;
+
+it('can calculate benin holidays', function () {
+    CarbonImmutable::setTestNowAndTimezone('2024-01-01');
```
Nielsvanpach, line 9:
```suggestion
    CarbonImmutable::setTestNow('2024-01-01');
```

composer.json
Massoud114: Can I extend the Turkey class to avoid repeating the long list of Islamic dates?

Nielsvanpach: As far as I understood @thecaliskan, this is different for most countries, so not really copy pastable.

tests/.pest/snapshots/Countries/BeninTest/it_can_calculate_benin_holidays.snap
Massoud114: The content of the snapshot is badly encoded for French characters (é,è,ê...), I don't know how to fix it.

Nielsvanpach:
Thanks!

PR #150, opened 2024-01-20
Add getInRange function

inDeev:
Added function to retrieve all holidays between two dates.

Description from docblock:
```
    /**
     * getInRange method allows you to pick holidays in a range of dates,
     *   - dates are inclusive.
     *   - dates are swappable, lower date could be passed as second argument.
     *   - dates could be a CarbonInterface or a string.
     *   - acceptable strings formats are 'Y-m-d' or 'Y-m' or 'Y'
     *   - if passed string is 'Y-m' or 'Y' it will be converted to first(from) / last{to} day of the month(from) / year(to)
     * E.g. to retrieve all holidays in between
     *    - 2020-01-01 and 2024-12-31, you could use: getInRange('2020-01-01', '2024-12-31'), getInRange('2020-01', '2024-12') or getInRange('2020', '2024')
     *    - 2024-06-01 and 2025-05-30, you could use: getInRange('2024-06-01', '2025-05-30'), getInRange('2024-06', '2025-05')
     */
```


src/Countries/Country.php
```diff
@@ -41,6 +41,41 @@ public function get(int $year, ?string $locale = null): array
         return $translatedHolidays;
     }
 
+    /** @return array<string, string>  date => name */
+    public function getInRange(?CarbonImmutable $from, ?CarbonImmutable $to): array
```
MirilTerolli, line 45: @inDeev IMO, the parameters should not be nullable. I can't think of a use case when someone might want to use this method and not pass both of the range parameters.
Accepting string parameters also might be useful. 
```suggestion
    public function getInRange(CarbonImmutable|string $from, CarbonImmutable|string $to): array
```

Chris53897:
@inDeev Thanks for this PR. That was just the function i am missing.

Any chance this is merged in the near future?
@Nielsvanpach 
@freekmurze

Nielsvanpach:
Thanks!

PR #152, opened 2024-01-21
Create IndiaTest.php

advit18:
Tests for India

Yes I am contributing for India


tests/Countries/IndiaTest.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Spatie\Holidays\Tests\Countries;
+
+use Carbon\CarbonImmutable;
+use Spatie\Holidays\Holidays;
```
dev-scripts, line 6: test and source code should be in a single PR. Better to combine both in one

PR #70, opened 2024-01-18
Added Turkey holidays

thecaliskan:

PR #155, opened 2024-01-21
Add Egyptian holidays

wessama:
Public holidays in Egypt, both observed and national holidays, including checks for whether or not a holiday falls on a weekday/weekend and when it should be observed. This is synced with the way Google Calendar does it.

As noted in #56 and #90, the use of `geniusts/hijri-dates` is necessary to (pretty much) accurately calculate when Islamic holidays will be observed during a given year. When they're observed differs from one country to another, even if they're all celebrated concurrently.

All tests pass.


wessama:
> Underlying package is abandoned and also doesn't have any tests -- would be very skeptical to depend on it.
> 
> It is also restricted to [Carbon v1 & v2](https://github.com/GeniusTS/hijri-dates/blob/af15f8849eea95b2faa463c751483be69bff37d2/composer.json#L14), which is preparing to have a new v3 release soon, so this package will be incompatible with Laravel 11: [laravel/framework#49764](https://github.com/laravel/framework/pull/49764)

Thanks. I'm working together with a few other PR authors to come up with a way to centralize this logic in the `Country` class.

wessama:
@Nielsvanpach I understand your hesitance to rely on an outdated package (one without tests, as well). So I adopted the Turkey approach in #70 and added those holidays statically. But I couldn't find a source that has been reliably tracking these holidays in Egypt prior to 2005. I double checked with Google Calendar and a few other sources, but the information on Wikipedia is incorrect.

You might know by now that there's no way to accurately predict when these holidays will be observed because of the variance involved, but this provides as good an approach as any. Maybe this can be maintained in the future to add any drastic changes. 

But apart from Islamic holidays, everything else can be precisely calculated.

I also found conflicts on the main branch with the `Egypt` class, which previously didn't account for Islamic holidays at all (and those actually factor into the fiscal year here and shouldn't be ignored, not by any reliable source of information anyway. I'd hate to be a business relying on this package only to find out it doesn't match real calendars). We've also been provided days off in lieu of national holidays that fall on a weekday or a weekend (which is Friday and Saturday here) for the past 3-4 years, this was also absent in the existing `Egypt` class.

Other than that, I found the existing lang translations a bit too Google translate-ish so I replaced them with more sensible, true-to-life translations and added missing ones:

- https://github.com/spatie/holidays/blob/c30b89bd97f29db8402f96653617593dcdfd3be5/lang/egypt/ar/holidays.json 

...and used the [Egyptian Presidency's calendar](https://www.presidency.eg/AR/%D9%85%D8%B5%D8%B1/%D8%A7%D9%84%D8%B9%D8%B7%D9%84%D8%A7%D8%AA-%D8%A7%D9%84%D8%B1%D8%B3%D9%85%D9%8A%D8%A9/) for reference.

Nielsvanpach:
Thanks!

PR #157, opened 2024-01-22
Albanian National Holidays

MirilTerolli:
Official Albanian holidays, as documented by the Albanian government portal [https://e-albania.al/Pages/OfficialHolidays.aspx](https://e-albania.al/Pages/OfficialHolidays.aspx)
Includes tentative dates for `Eid al-Fitr` and `Eid al-Adha` holidays until `2034`, because the dates are not static.

### DIFF
--- a/src/Countries/Albania.php
+++ b/src/Countries/Albania.php
@@ -26,7 +26,7 @@ class Albania extends Country
         ], $this->variableHolidays($year));
     }
 
-    /** @return array<string, CarbonImmutable> */
+    /** @return array<string, CarbonImmutable|string> */
     protected function variableHolidays(int $year): array
     {
         return [
@@ -40,13 +40,13 @@ class Albania extends Country
     /**
      * 
      */
-    private function getEidAlFitrHoliday(int $year): CarbonImmutable
+    private function getEidAlFitrHoliday(int $year): string
     {
         /**
          * Provided until 2034 by qppstudio.net.
          * https://www.qppstudio.net/global-holidays-observances/eid-al-fitr-end-of-ramadan.htm
          */
-        $date = match ($year) {
+        return match ($year) {
             2024 => '04-10',
             2025 => '03-30',
             2026 => '03-20',
@@ -60,19 +60,16 @@ class Albania extends Country
             2034 => '12-12',
             default => '01-01' // Temporary placeholder; requires ongoing maintenance.
         };
-
-        return CarbonImmutable::createFromFormat('Y-m-d', "$year-$date")
-            ->startOfDay();
     }
 
-    private function getEidAlAdhaHoliday(int $year): CarbonImmutable
+    private function getEidAlAdhaHoliday(int $year): string
     {
         /**
          * Tentative dates.
          * Provided until 2034 by timeanddate.com.
          * https://www.timeanddate.com/holidays/us/eid-al-adha
          */
-        $date = match ($year) {
+        return match ($year) {
             2024 => '06-17',
             2025 => '06-07',
             2026 => '05-27',
@@ -86,9 +83,6 @@ class Albania extends Country
             2034 => '03-01',
             default => '01-01' // Temporary placeholder; requires ongoing maintenance.
         };
-
-        return CarbonImmutable::createFromFormat('Y-m-d', "$year-$date")
-            ->startOfDay();
     }
 
 }
--- a/src/Countries/Albania.php
+++ b/src/Countries/Albania.php
@@ -1,0 +1,94 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\CarbonImmutable;
+
+class Albania extends Country
+{
+    public function countryCode(): string
+    {
+        return 'al';
+    }
+
+    protected function allHolidays(int $year): array
+    {
+        return array_merge([
+            'Viti i Ri' => '01-01',
+            'Dita e Verës' => '03-14',
+            'Dita e Sulltan Nevruzit' => '03-22',
+            'Dita Ndërkombëtare e Punëtorëve' => '05-01',
+            'Dita e Shenjtërimit të Shenjt Terezës' => '09-05',
+            'Dita e Pavarësisë' => '11-28',
+            'Dita e Çlirimit' => '11-29',
+            'Dita Kombëtare e Rinisë' => '12-08',
+            'Krishtlindjet' => '12-25',
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        return [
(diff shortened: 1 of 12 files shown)


src/Countries/Albania.php
```diff
@@ -0,0 +1,88 @@
...
+        return array_merge([
+            'Viti i Ri' => '01-01',
+            'Dita e Verës' => '03-14',
+            'Dita e Sulltan Nevruzit' => '03-22',
+            'Dita Ndërkombëtare e Punëtorëve' => '05-01',
+            'Dita e Shenjtërimit të Shenjt Terezës' => '09-05',
+            'Dita e Pavarësisë' => '11-28',
+            'Dita e Çlirimit' => '11-29',
+            'Dita Kombëtare e Rinisë' => '12-08',
+            'Krishtlindjet' => '12-25',
```
fitimvata, line 25: @MirilTerolli `Krishtlindjet` should be `Krishtlindja`

fitimvata (changes requested):

src/Countries/Albania.php
MirilTerolli: @fitimvata it's done.

Nielsvanpach:
In case you want to add multi language support: https://github.com/spatie/holidays/tree/main/lang

MirilTerolli:
> In case you want to add multi language support: https://github.com/spatie/holidays/tree/main/lang

Good idea @Nielsvanpach , it's done! 👍

src/Countries/Albania.php
```diff
@@ -0,0 +1,88 @@
...
+            2026 => '03-20',
+            2027 => '03-09',
+            2028 => '02-26',
+            2029 => '02-14',
+            2030 => '02-04',
+            2031 => '01-24',
+            2032 => '01-14',
+            2033 => '01-02',
+            2034 => '12-12',
+            default => '01-01' // Temporary placeholder; requires ongoing maintenance.
```
Nielsvanpach, line 61: I would throw an exception by default and make sure we don't return false data

src/Countries/Albania.php
```diff
@@ -0,0 +1,88 @@
...
+            2026 => '05-27',
+            2027 => '05-17',
+            2028 => '05-05',
+            2029 => '04-24',
+            2030 => '04-14',
+            2031 => '04-03',
+            2032 => '03-22',
+            2033 => '03-12',
+            2034 => '03-01',
+            default => '01-01' // Temporary placeholder; requires ongoing maintenance.
```
Nielsvanpach, line 84: I would throw an exception by default and make sure we don't return false data

src/Countries/Albania.php
MirilTerolli: @Nielsvanpach , that would be a good idea, but this would prevent other standard holidays from not being displayed in case if we run something like this ``Holidays::for(country: "al", year: 2035)->get();`` 🤷‍♂️

src/Countries/Albania.php
MirilTerolli: @Nielsvanpach as mentioned above, throwing an exception will prevent the standard holidays from being displayed. 
I wonder if this is what we want.

Nielsvanpach: If we can't provide valid data, I prefer to throw an exception.

src/Countries/Albania.php
```diff
@@ -0,0 +1,88 @@
...
+            'E diela e Pashkëve Ortodokse' => $this->orthodoxEaster($year),
+            'Dita e Bajramit të Madh' => $this->getEidAlFitrHoliday($year),
+            'Dita e Kurban Bajramit' => $this->getEidAlAdhaHoliday($year),
+        ];
+    }
+
+    /**
+     * 
+     */
+    private function getEidAlFitrHoliday(int $year): string
```
fitimvata, line 43:
```suggestion
    private function getEidAlFitrHoliday(int $year): ?string
```

src/Countries/Albania.php
fitimvata, line 61:
```suggestion
            default => null // Temporary placeholder; requires ongoing maintenance.
```

src/Countries/Albania.php
```diff
@@ -0,0 +1,88 @@
...
+            2030 => '02-04',
+            2031 => '01-24',
+            2032 => '01-14',
+            2033 => '01-02',
+            2034 => '12-12',
+            default => '01-01' // Temporary placeholder; requires ongoing maintenance.
+        };
+    }
+
+    private function getEidAlAdhaHoliday(int $year): string
```
fitimvata, line 65:
```suggestion
    private function getEidAlAdhaHoliday(int $year): ?string
```

src/Countries/Albania.php
fitimvata, line 84:
```suggestion
            default => null // Temporary placeholder; requires ongoing maintenance.
```

src/Countries/Albania.php
```diff
@@ -0,0 +1,88 @@
...
+            'Dita e Çlirimit' => '11-29',
+            'Dita Kombëtare e Rinisë' => '12-08',
+            'Krishtlindja' => '12-25',
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable|string> */
+    protected function variableHolidays(int $year): array
+    {
+        return [
```
fitimvata, line 32:
```suggestion
        return array_filter([
```

src/Countries/Albania.php
```diff
@@ -0,0 +1,88 @@
...
+            'E diela e Pashkëve Katolike' => $this->easter($year),
+            'E diela e Pashkëve Ortodokse' => $this->orthodoxEaster($year),
+            'Dita e Bajramit të Madh' => $this->getEidAlFitrHoliday($year),
+            'Dita e Kurban Bajramit' => $this->getEidAlAdhaHoliday($year),
+        ];
```
fitimvata, line 37:
```suggestion
        ]);
```

fitimvata (changes requested):
@MirilTerolli I personally would do something like requested changes. Return null from `getEidAlFitrHoliday` and `getEidAlAdhaHoliday` and filter array in `variableHolidays`

src/Countries/Albania.php
MirilTerolli: @fitimvata that could work as well, changes pushed.

src/Countries/Albania.php
MirilTerolli: @fitimvata pushed.

src/Countries/Albania.php
MirilTerolli: @fitimvata pushed.

src/Countries/Albania.php
MirilTerolli: @fitimvata changes pushed.

fitimvata (approved):

PR #160, opened 2024-01-22
Add Japan support

mkokio:

1. Create a new class in the Countries directory. It should extend the Country class. Done.
2. Add a test for the new country in the tests directory. Done.
3. Run the tests so a snapshot gets created. Done.
4. Verify the result in the newly created snapshot is correct. Done.


Nielsvanpach:
Thanks!

PR #168, opened 2024-01-23
Provide optional parameters to Countries

inDeev:
This PR allow to provide any required parameters to countries models like region, ethnical type, ...

This PR was created by discussion about North Macedonia holidays [Thread here](https://github.com/spatie/holidays/discussions/104) so NorthMacedonia has some preparations made, which will be updated by its contributor

This PR also updating French model and test, as there was problem with custom France constructor just to use region (now it uses global one.


Nielsvanpach:
I don't think public holidays are based on someone's ethnic?

Nielsvanpach:
I don't really like this approach, so closing this one. Thanks for your work anyway!

PetrKaterinak:
@Nielsvanpach Ethnic group is only one of possibilities of this PR, and if it is given by government as discussed in thread mentioned in original PR message, it is valid to have ethnical based holidays.

PR #171, opened 2024-01-23
Add Montenegro holidays

Rkishahaha:
Using data from: https://neradni-dani.com/neradni-dani-praznici-2024-cg.php


Nielsvanpach:
In case you want to add multi language support: https://github.com/spatie/holidays/tree/main/lang

Nielsvanpach:
Thanks!

PR #176, opened 2024-01-24
Add Brazil regional holidays

ezequidias:
Implementing state holidays in Brazil


Nielsvanpach (changes requested):
Could you add a dedicated test for this, run this locally and verify if the generated snapshot results are correct?

PR #177, opened 2024-01-24
Add support for multi languages 

Nielsvanpach:


tests/HolidaysTest.php
```diff
@@ -91,3 +91,9 @@
     $result = Holidays::has(country: 'unknown');
     expect($result)->toBeFalse();
 });
+
+it('can get translated holiday names', function () {
+    $result = Holidays::for(country: 'be', year: 2020, locale: 'fr')->get();
+
+    expect(formatDates($result))->toMatchSnapshot();
```
freekmurze, line 98: We don't have to verify everything in the result, only that we get a translated name. I would use a specific assertion instead of a snapshot one

freekmurze (changes requested):
Looks good!

herpaderpaldent:
Thank you for this addition. This will help certainly!

However i have one question: 

The current solution with language works via the 'magic' method but not if you would create the Country Object as you would do in the current README. 

https://github.com/spatie/holidays/blob/faf22be7976c1f56505ba4aa69447db81c212db9/README.md?plain=1#L119-L148

This might be because https://github.com/spatie/holidays/blob/faf22be7976c1f56505ba4aa69447db81c212db9/src/Countries/Country.php#L44 

does not accept named parameters. Are you open to accept named parameter in that method and drop `...func_get_args()` as well? 

Or is the construction of holidays via the magic method the standard?

Nielsvanpach:
I'm not sure I understand what you mean. Could you add a PR with a failing test?

I've included a test on how to use this with named params: 
https://github.com/spatie/holidays/pull/177/files#diff-198ea683244cb0b1bcd6bc0093bf6a397a011ee9dcbbaca1e15917f6acbf54bcR96

PR #182, opened 2024-01-26
feat: Add Bangla language support for Bangladeshi holidays

jfBiswajit:


Nielsvanpach (approved):
Thanks. Does it make sense to return Bangla by default instead of English?

jfBiswajit:
"Thanks! Please check the changes. It returns Bangla names in English instead of Bangla font.

Nielsvanpach:
Thanks!

PR #183, opened 2024-01-27
Germany: Adding local for EN, FR, NL

arnebr:
* Add Germany holidays local support for: EN, FR, NL
* Add tests for German holidays with multiple locales and unsupported locale exception

### DIFF
--- a/tests/Countries/GermanyTest.php
+++ b/tests/Countries/GermanyTest.php
@@ -4,6 +4,7 @@ namespace Spatie\Holidays\Tests\Countries;
 
 use Carbon\CarbonImmutable;
 use Spatie\Holidays\Countries\Germany;
+use Spatie\Holidays\Exceptions\InvalidLocale;
 use Spatie\Holidays\Holidays;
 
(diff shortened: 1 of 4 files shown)


tests/Countries/GermanyTest.php
```diff
@@ -78,3 +79,20 @@
...
+})->with(
+    [
+        ['en','New Year'],
+        ['nl','Nieuwjaar'],
+        ['fr','Jour de l\'An']
+    ]
+);
+it('cannot get translated holiday names for unsupported locales', function () {
+    Holidays::for(country: 'de', year: 2024, locale: 'xx')->get();
+})->throws(InvalidLocale::class, 'Locale `xx` is not supported for country `Germany`.');
```
Nielsvanpach, line 98:
```suggestion
```

Nielsvanpach: 👍 for the test, but it's duplicate test case from another class

Nielsvanpach (commented):
Thanks!

PR #184, opened 2024-01-27
Fixing README.md to update region parameter in country constructor to a working example

arnebr:
Before the README.md region settings has been mixed with two different countries.
```php
$holidays = Holidays::for(Austria::make(region: 'de-bw'))->get();
```
Austria has called a German region. I´ve adjusted it to the German working example and uppercased the region parameter as it is in the German class.


Nielsvanpach:
Thanks!

PR #187, opened 2024-01-28
Add Bahrain 

bo3bdo:


Nielsvanpach:
There has been some discussion about using that library.
Here is why we haven't used it: https://github.com/spatie/holidays/blob/main/src/Countries/Turkey.php#L11

bo3bdo:
ok i update the code

src/Countries/Bahrain.php
```diff
@@ -0,0 +1,30 @@
...
+    public function countryCode(): string
+    {
+        return 'bh';
+    }
+
+    protected function allHolidays(int $year): array
+    {
+        return array_merge([
+            'New Year Day' => '1-1',
+            'Labour Day' => '5-1',
```
Nielsvanpach, line 19: This PR lacks holidays. Is it possible to extend with all public holidays? For islamic holidays, you could look at how Turkey or Egypt did the implementation: https://github.com/spatie/holidays/pull/70

PR #188, opened 2024-01-28
Germany: Setting some historical holidays

arnebr:
Added:

- [Uprising Day](https://en.wikipedia.org/wiki/East_German_uprising_of_1953) (17 June) This day was public holiday under the title of "German Unity Day" from 1954 until 1990 when that unity actually was achieved.
- Saint Joseph's Day was a holidays in Bavaria after WW2 until 1969
- In 2020, a regional holiday in Berlin occurred on 8 May to mark the [75th anniversary of surrender](https://en.wikipedia.org/wiki/Victory_in_Europe_Day)
- Added the translations for the dates above

Fixed:
- Test case for buß und bettag was set to reformationstag

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,13 +128,13 @@ In case your country has specific rules for calculating holidays,
 for example region specific holidays, you can pass this to the constructor of your country class.
 
 ```php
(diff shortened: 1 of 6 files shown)


Nielsvanpach:
Great! Thanks

PR #196, opened 2024-01-30
Clean Tests

michaelnabil230:


Jampire:
Why do we need to rebuild tests from one style to another?

Nielsvanpach:
I don't think this is needed.

PR #137, opened 2024-01-19
Add holidays for Sweden

robost:
Re-opening [https://github.com/spatie/holidays/pull/31](https://github.com/spatie/holidays/pull/31) as it was closed by mistake.

PR #199, opened 2024-02-01
Add support for Swedish holidays

adevade:
There's already a [pull request open for Sweden](https://github.com/spatie/holidays/pull/137). This PR fixes the requested changes.

### DIFF
--- a/src/Countries/Sweden.php
+++ b/src/Countries/Sweden.php
@@ -26,17 +26,17 @@ class Sweden extends Country
     /** @return array<string, CarbonImmutable> */
     protected function variableHolidays(int $year): array
     {
-        $easter = $this->easter($year)->setTimezone('Europe/Stockholm');
+        $easter = $this->easter($year);
 
         /** @var CarbonImmutable $midsummerDay */
-        $midsummerDay = CarbonImmutable::createFromFormat('Y-m-d', "{$year}-06-20", 'Europe/Stockholm');
+        $midsummerDay = CarbonImmutable::createFromFormat('Y-m-d', "{$year}-06-20");
 
         if (! $midsummerDay->isSaturday()) {
             $midsummerDay = $midsummerDay->next(CarbonImmutable::SATURDAY);
         }
 
         /** @var CarbonImmutable $halloween */
-        $halloween = CarbonImmutable::createFromFormat('Y-m-d', "{$year}-10-31", 'Europe/Stockholm');
+        $halloween = CarbonImmutable::createFromFormat('Y-m-d', "{$year}-10-31");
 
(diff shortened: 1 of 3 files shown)


src/Countries/Sweden.php
```diff
@@ -0,0 +1,57 @@
...
+            'Juldagen' => '12-25',
+            'Annandag jul' => '12-26',
+        ], $this->variableHolidays($year));
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        $easter = $this->easter($year)->setTimezone('Europe/Stockholm');
+
```
Nielsvanpach, line 30:
```suggestion
        $easter = $this->easter($year);

```

Nielsvanpach: Timezones shouldn't be needed

src/Countries/Sweden.php
```diff
@@ -0,0 +1,57 @@
...
+        /** @var CarbonImmutable $midsummerDay */
+        $midsummerDay = CarbonImmutable::createFromFormat('Y-m-d', "{$year}-06-20", 'Europe/Stockholm');
```
Nielsvanpach, line 32: Timezones shouldn't be needed. You can call ->startOfDay() on the Carbon instance.

Example: https://github.com/spatie/holidays/blob/main/src/Countries/Country.php#L52

src/Countries/Sweden.php
```diff
@@ -0,0 +1,57 @@
...
+
+        if (! $midsummerDay->isSaturday()) {
+            $midsummerDay = $midsummerDay->next(CarbonImmutable::SATURDAY);
+        }
+
+        /** @var CarbonImmutable $halloween */
+        $halloween = CarbonImmutable::createFromFormat('Y-m-d', "{$year}-10-31", 'Europe/Stockholm');
```
Nielsvanpach, line 39: Same comment as above

Nielsvanpach (changes requested):
Thanks! In case you need translations: https://github.com/spatie/holidays/tree/main/lang

src/Countries/Sweden.php
adevade: I saw that some countries added timezones, and some do not. 

https://github.com/spatie/holidays/blob/914ae68e4dd979172182fb60058ae914db12c0f8/src/Countries/Australia.php#L33

Nielsvanpach: Thanks, I fixed some of them in the meantime.

Nielsvanpach (approved):
Thanks!

PR #203, opened 2024-02-06
Add holidays for Bosnia And Herzegovina

DesaBL:

### DIFF
--- a/src/Countries/BosniaAndHerzegovina.php
+++ b/src/Countries/BosniaAndHerzegovina.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace Spatie\Holidays\Countries;
+
+use Carbon\CarbonImmutable;
+
+class BosniaAndHerzegovina extends Country
+{
+    protected function __construct(
+        protected ?string $region = null,
+    ) {
+    }
+
+    public function countryCode(): string
+    {
+        return 'ba';
+    }
+
+    protected function allHolidays(int $year): array
+    {
+        return array_merge([
+            'Nova godina - prvi dan' => '01-01',
+            'Nova godina - drugi dan' => '01-02',
+            'Badnji dan (za pravoslavce)' => '01-06',
+            'Božić (za pravoslavce)' => '01-07',
+            'Praznik rada - prvi dan' => '05-01',
+            'Praznik rada - drugi dan' => '05-02',
+            'Badnji dan (za rimokatolike)' => '12-24',
+            'Božić (za rimokatolike)' => '12-25',
+        ], $this->regionHolidays());
+    }
+
+    /** @return array<string, string> */
+    protected function regionHolidays(): array
+    {
+        switch ($this->region) {
+            case 'BA-RS':
+                return [
+                    'Dan Republike' => '01-09',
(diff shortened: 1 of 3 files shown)


src/Countries/BosniaAndHerzegovina.php
```diff
@@ -0,0 +1,83 @@
...
+
+    public function countryCode(): string
+    {
+        return 'ba';
+    }
+
+    protected function allHolidays(int $year): array
+    {
+        return array_merge([
+            'Nova godina - prvi dan' => '01-01',
```
Nielsvanpach, line 22: Translations needed?
https://github.com/spatie/holidays/tree/main/lang

src/Countries/BosniaAndHerzegovina.php
```diff
@@ -0,0 +1,83 @@
...
+            'Praznik rada - drugi dan' => '05-02',
+            'Badnji dan (za rimokatolike)' => '12-24',
+            'Božić (za rimokatolike)' => '12-25',
+        ], $this->regionHolidays());
+    }
+
+    /** @return array<string, string> */
+    protected function regionHolidays(): array
+    {
+        switch ($this->region) {
```
Nielsvanpach, line 36: Can you use a match() here instead of switch()?

src/Countries/BosniaAndHerzegovina.php
```diff
@@ -0,0 +1,83 @@
...
+            case 'BA-RS':
```
Nielsvanpach, line 37: I prefer to lowercase the incoming region variable and use lowercased cases

src/Countries/BosniaAndHerzegovina.php
```diff
@@ -0,0 +1,83 @@
...
+                ];
+        }
+
+        return [];
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    public function variableHolidays(int $year): array
+    {
+        // Orthodox Easter calculation needs to be in the same timezone as the country
```
Nielsvanpach, line 60: What do you mean with this comment? Is `$this->orthodoxEaster($year)` giving a bad result for you?

tests/Countries/BosniaAndHerzegovinaTest.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Spatie\Holidays\Tests\Countries;
+
+use Carbon\CarbonImmutable;
+use Spatie\Holidays\Holidays;
+
+it('can calculate bosnia and herzegovina holidays', function () {
```
Nielsvanpach, line 8: Could you also add a test case with a region?

Nielsvanpach (changes requested):

src/Countries/BosniaAndHerzegovina.php
DesaBL: @Nielsvanpach I take a look at some of the already implemented countries (Germany, to be more precise) and I saw they used uppercases there, so that's the reason why I decided to use uppercase

Nielsvanpach (approved):
Thanks!

PR #207, opened 2024-02-18
Add New Zealand Holidays

jpgnz:
Adding New Zealand holidays. Took guidance from Australia and Wales.

### DIFF
--- a/tests/Countries/NewZealandTest.php
+++ b/tests/Countries/NewZealandTest.php
@@ -1,0 +1,137 @@
+<?php
+
+use Carbon\CarbonImmutable;
+use Spatie\Holidays\Holidays;
+
+it('calculates new zealand holidays by date', function (string $name, string $date) {
+
+    $testDate = CarbonImmutable::parse($date);
+
+    $holidayName = Holidays::for('nz')->getName($testDate);
+    $isHoliday = Holidays::for('nz')->isHoliday($testDate);
+
+    expect($holidayName)->toBe($name)->and($isHoliday)->toBeTrue();
+
+})->with([
+    //2023
+    ['New Year\'s Day (Mondayisation)', '2023-01-02'],
+    ['Day after New Year\'s Day (Mondayisation)', '2023-01-03'],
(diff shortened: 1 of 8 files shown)


tests/Countries/NewZealandTest.php
```diff
@@ -0,0 +1,137 @@
...
+
+});
+
+it('returns a substitute day if new years day falls on a weekend', function () {
+    CarbonImmutable::setTestNow('2033-01-01');
+
+    $holidays = Holidays::for(country: 'nz')->get();
+
+    expect($holidays)->toBeArray()->not()->toBeEmpty();
+    expect(formatDates($holidays))->toMatchSnapshot();
```
Nielsvanpach, line 100: Instead of a complete snapshot test, it might be more interesting to only test the specific case (it has a substitute day for new year)

tests/Countries/NewZealandTest.php
```diff
@@ -0,0 +1,137 @@
...
+});
+
+it('can calculate new zealand holidays if christmas is on a friday', function () {
+    CarbonImmutable::setTestNow('2028-01-01');
+
+    $holidays = Holidays::for(country: 'nz')->get();
+
+    expect($holidays)->toBeArray()->not()->toBeEmpty();
+    expect(formatDates($holidays))->toMatchSnapshot();
```
Nielsvanpach, line 109: Instead of a complete snapshot test, it might be more interesting to only test the specific case

Nielsvanpach (commented):
Thanks!

tests/Countries/NewZealandTest.php
jpgnz: I took guidance from [WalesTest.php](https://github.com/spatie/holidays/blob/main/tests/Countries/WalesTest.php) as I wasn't quite sure what coverage would be appropriate.

Nielsvanpach: That test had been better with specific assertions too

PR #209, opened 2024-02-24
Add Jordan holidays

yousefdev20:


src/Calendars/Islamic/arafat-holidays.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+return [
+
+    'jo' => [
```
Nielsvanpach, line 5: Can you use class consts for this? Have a look at Egypt or Bahrain

src/Concerns/IslamicHolidays.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Spatie\Holidays\Concerns;
+
+use Carbon\CarbonImmutable;
+
+trait IslamicHolidays
```
Nielsvanpach, line 7: We already have an IslamicCalendar trait which you can use

src/Concerns/WeekendHolidays.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Spatie\Holidays\Concerns;
+
+use RuntimeException;
+use Carbon\CarbonPeriod;
+use Carbon\CarbonImmutable;
+
+trait WeekendHolidays
```
Nielsvanpach, line 9: I think the Observable trait is what you need

Nielsvanpach (changes requested):

src/Calendars/Islamic/arafat-holidays.php
yousefdev20: @Nielsvanpach 
Why I used this approach:

1. It's hard to read around 200 lines of code to read the elements.
2. in some countries the holidays are shared or in the same region.
3. at the end the Islamic directory is loadable.

src/Concerns/IslamicHolidays.php
yousefdev20: This trits depends on the constants variable and I explained the issue in the above replay.

src/Calendars/Islamic/arafat-holidays.php
Nielsvanpach: 2 and 3 seem valid points.

Although, using `jo` as a key doesn't feel like the way to go for being shareable across countries?

Also, if you have a look at Turkey, I prefer that format. What the benefit of using the extra `date` key? 

Also, for multiday holidays it's possible to have it twice in a single year:

```php
1999 => '01-19',
2000 => [
    '01-08',
     '12-27',
],
```

yousefdev20: `jo` it may change,

`date` is more readable,

The third note is valid.

PR #211, opened 2024-03-07
Fix: Canada Victoria Day correction in test and calculation

mercury64:
Corrected the Victoria Day date for 2024 to be May 20th in test snapshot, not May 27th.

Corrected the code in src/Countries/Canada.php to calculate the correct date for Victoria Day.

Canada Victoria Day is always the first Monday in May preceding May 25th(and not including).
The Monday will always be between May 18th and May 24th.


Nielsvanpach:
Thanks!

PR #217, opened 2024-04-02
Fix the Holidays in French speaking Switzerland

OlinB:
I dug through the various French speaking Cantons' legislation to implement their holidays correctly


Nielsvanpach:
Thanks!

PR #218, opened 2024-04-09
Add holidays for Ecuador

insoutt:

My second commit is related following the rules of #219


src/Countries/Ecuador.php
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        return $date;
+    }
+
+    public function getChristmasHoliday(int $year)
+    {
+        $date = CarbonImmutable::createFromDate($year, 12, 25);
+
+        if($year === 2022) {
```
Nielsvanpach, line 44: Do you need this behaviour?
https://github.com/spatie/holidays/blob/main/src/Concerns/Observable.php#L36

src/Countries/Ecuador.php
```diff
@@ -0,0 +1,72 @@
...
+            return $date->addDay();
+        }
+
+        return $date;
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        $easter = $this->easter($year);
```
Nielsvanpach, line 54: Instead of mixing two languages, I suggest to use English here and implement `HasTranslations` & `Translatable`.

Nielsvanpach (changes requested):
Thanks!

src/Countries/Ecuador.php
insoutt: Yes, I already implemented it

src/Countries/Ecuador.php
insoutt: ok, I just made the changes to use translations

src/Countries/Ecuador.php
```diff
@@ -0,0 +1,81 @@
...
+    use Translatable;
+
+    public function countryCode(): string
+    {
+        return 'ec';
+    }
+
+    public function defaultLocale(): string
+    {
+        return 'en';
```
Nielsvanpach, line 22:
```suggestion
        return 'es;
```

src/Countries/Ecuador.php
```diff
@@ -0,0 +1,81 @@
...
+        if($date->is('Wednesday') || $date->is('Thursday')) {
+            return $date->next(CarbonImmutable::FRIDAY);
+        }
+
+        return $date;
+    }
+
+    public function getChristmasHoliday(int $year): CarbonImmutable
+    {
+        if($year === 2022) {
```
Nielsvanpach, line 53: Why is only 2022 an observed christmas?

src/Countries/Ecuador.php
```diff
@@ -0,0 +1,81 @@
...
+        return CarbonImmutable::createFromDate($year, 12, 25);
+    }
+
+    /** @return array<string, CarbonImmutable> */
+    protected function variableHolidays(int $year): array
+    {
+        $easter = $this->easter($year);
+        $ashWednesday = $easter->subDays(46);
+        $carnivalMonday = $ashWednesday->subDays(2);
+        $carnivalTuesday = $ashWednesday->subDay();
```
Nielsvanpach, line 66:
```suggestion
```

tests/Countries/EcuadorTest.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Spatie\Holidays\Tests\Countries;
+
+use Carbon\CarbonImmutable;
+use Spatie\Holidays\Holidays;
+
+it('can calculate ecuador holidays', function () {
```
Nielsvanpach, line 8: Can you also add a test to cover the translation?

src/Countries/Ecuador.php
```diff
@@ -0,0 +1,81 @@
...
+
+        return [
+            'Holy Friday' => $easter->subDays(2),
+            'Carnival Monday' => $carnivalMonday,
+            'Carnival Tuesday' => $carnivalTuesday,
```
Nielsvanpach, line 71:
```suggestion
            'Carnival Monday' => $ashWednesday->subDays(2),
            'Carnival Tuesday' => $ashWednesday->subDay(),
```

src/Countries/Ecuador.php
insoutt: Because only in 2022 Christmas was moved from Sunday to Monday, and also I made a mistake using `observedChristmasDay` because it changes from Sunday to Tuesday, so I changed to use `sundayToNextMonday` [d693203](https://github.com/spatie/holidays/pull/218/commits/d693203c0a7695b1c2824afdf56bf8a6b421e8d1)

tests/Countries/EcuadorTest.php
insoutt: Sure, but I had to change `defaultLocale()` to `en` to test Spanish and English translations, If I set the `defaultLocale()` to `es` I cannot test the English translations

Nielsvanpach: By running `./vendor/bin/pest --update-snapshots` you can update the snapshots. Just make sure the generated results are correct.

insoutt: Sure I just did it, the snapshots are correct now.

Nielsvanpach (approved):
Thanks!

PR #224, opened 2024-04-29
Update Sweden.php to always use leading zero in day

adevade:
Just adding missing zeroes to two dates for consistency.


Nielsvanpach:
Thanks!

PR #237, opened 2024-09-13
Add New Country (Malaysia)

fathikhalid7:

1. Create a new class in the Countries directory. It should extend the Country class. CHECKED
2. Add a test for the new country in the tests directory. CHECKED
3. Run the tests so a snapshot gets created. CHECKED
4. Verify the result in the newly created snapshot is correct. CHECKED


Nielsvanpach:
Malaysia is already merged. Feel free to send an additional PR in case something is not working as expected.

PR #255, opened 2024-12-19
Fix: some of the dates for Argentina where incorrectly formated

Akrista:
Fix: some of the dates for Argentina where incorrectly formated


Nielsvanpach:
Thanks!

Nielsvanpach:
Could you fix the tests too?

PR #265, opened 2025-04-25
refactor: simplify null coalescing and optional chaining in getOverlapping

Ayoub-Mabrouk:
Replaced explicit null check and exception throw with null coalescing + throw expression.
Also added optional chaining to ensure safer date calculation using CarbonImmutable.


Nielsvanpach:
Thanks

Ayoub-Mabrouk:
> Thanks

Your welcome.

PR #266, opened 2025-04-25
perf: implement lazy-initialized Chinese calendar formatter

Ayoub-Mabrouk:
- Formatter initializes only upon first use
- Caches formatter for subsequent calls
- Invalidation handled on timezone changes


Nielsvanpach:
Thanks!

PR #270, opened 2025-04-30
fix: remove non-variable holiday "Cambio de Gobierno" from Mexico

### LINKED ISSUE #262: [Bug]: Mexico "Cambio de Gobierno" holiday is added in every year
### What happened?

When retrieving Mexico’s holidays, the function always returns the "Cambio de Gobierno" holiday, even though it is classified as a variable holiday (it occurs every 6 years).

The governmentChangeDate function in Mexico.php works correctly, but the date is still being added in the allHolidays function as follows:

   protected function allHolidays(int $year): array
    {
        return array_merge([
            'Año Nuevo' => '01-01',
            'Día Internacional de los Trabajadores' => '05-01',
            'Día de Independencia' => '09-16',
            'Cambio de Gobierno' => '10-01', // ---> Here
            'Navidad' => '12-25',
        ], $this->variableHolidays($year));
    }

### How to reproduce the bug

Try to get holidays of Mexico with any year.

$holidays = Holidays::for(country: 'mx')->get();

### Package Version

1.17.0

### PHP Version

8.3.16

### Which operating systems does with happen with?

_No response_

### Notes

_No response_

Kenny1291:
Fixes #262

### DIFF
--- a/src/Countries/Mexico.php
+++ b/src/Countries/Mexico.php
@@ -17,7 +17,6 @@ class Mexico extends Country
             'Año Nuevo' => '01-01',
             'Día Internacional de los Trabajadores' => '05-01',
             'Día de Independencia' => '09-16',
-            'Cambio de Gobierno' => '10-01',
             'Navidad' => '12-25',
         ], $this->variableHolidays($year));
     }


Nielsvanpach:
Thanks!

PR #275, opened 2025-05-14
Fixed Turkey.php's religious holidays in line with the official site of the Presidency of Religious Affairs

deniztezcan:
Inline with the official list of Diyanet found at vakithesaplama.diyanet.gov.tr


Nielsvanpach:
Could you update the snapshots in the tests too?

PR #328, opened 2026-04-15
fix: add Easter Sunday and use official Epiphany name for Croatia

vblazenka:
Fixes two issues with Croatian public holidays:
  - **Missing** `Uskrs` (Easter Sunday) -> listed in the Public Holidays Act and on the source of truth, but absent from `Croatia.php`.
  - **Wrong name** for Epiphany: `Bogojavljenje` → `Bogojavljenje ili Sveta tri kralja`, matching the official name in *Zakon o blagdanima, spomendanima i neradnim danima u Republici Hrvatskoj*.

  Source of truth: https://neradni-dani.com/neradni-dani-2026.php

### DIFF
--- a/src/Countries/Croatia.php
+++ b/src/Countries/Croatia.php
@@ -15,7 +15,7 @@ class Croatia extends Country
     {
         return array_merge([
             Holiday::national('Nova godina', "{$year}-01-01"),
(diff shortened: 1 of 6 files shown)


Nielsvanpach:
Thanks!

PR #339, opened 2026-08-05
Add Hong Kong Public Holidays

littpi:


freekmurze (changes requested):
Thanks for putting this together, @littpi. The observed/shifting logic and the Lunar New Year day-naming are nicely done. A few things need fixing before this can go in.

### 1. Ching Ming Festival is hardcoded to 5 April

```php
'清明節' => "{$year}-04-05",
```

Ching Ming follows the Qingming solar term, which alternates between 4 and 5 April. The gazetted HK dates are:

| Year | Actual | This PR |
| --- | --- | --- |
| 2024 | 4 Apr | 5 Apr |
| 2025 | 4 Apr | 5 Apr |
| 2026 | 5 Apr | 5 Apr |
| 2027 | 5 Apr | 5 Apr |
| 2028 | 4 Apr | 5 Apr |
| 2029 | 4 Apr | 5 Apr |

So roughly half the years are wrong, including the snapshot year. The committed snapshot currently bakes in `清明節 => 2024-04-05`, which should be `2024-04-04`.

This also cascades: in 2026 the wrong-by-luck-correct date happens to be a Sunday and gets shifted, but in years where the real date differs the observed shifts will be off too.

Other countries in this package handle non-computable dates with a year-keyed lookup map (see `India.php`, `Benin.php`, `Ethiopia.php`). That pattern would work well here.

### 2. Missing translations

Several names generated by `observedHolidayName()` have no entry in `lang/hk/en/holidays.json`:

- `勞動節翌日` (2022, 2033)
- `聖誕節翌日` (2022, 2033)
- `一月一日翌日` (2023, 2034)
- `國慶日翌日` (2023, 2028, 2034)
- `香港特別行政區成立紀念日翌日` (2029, 2035)

Conversely, `清明節翌日` and `復活節星期一翌日` are in the JSON but `復活節星期一翌日` can never be produced (Easter Monday is never a Sunday).

### 3. Duplicate holiday names on consecutive days

When a holiday whose name already contains `翌日`, or `聖誕節後第一個周日`, lands on a Sunday, the shifted holiday reuses the exact same name:

```
2021-12-26 Sun  聖誕節後第一個周日
2021-12-27 Mon  聖誕節後第一個周日   <- observed
2022-09-11 Sun  中秋節翌日
2022-09-12 Mon  中秋節翌日           <- observed
```

Officially there is only one holiday in each case (27 Dec 2021, 12 Sep 2022). Related: `Jamaica.php` replaces the original with the observed one rather than emitting both. Worth deciding whether HK should do the same, given the original entries here are always Sundays.

### 4. Null lunar dates are cast to an empty string

```php
$holidayObjects[] = Holiday::national($name, (string) $date);
```

`lunarHolidays()` is typed `array<string, string|null>`, and `(string) null` is `''`, which `CarbonImmutable::parse()` resolves to *now*. If the ICU conversion ever fails you get a silent bogus holiday on today's date. `Taiwan.php` skips nulls instead.

---

Not a blocker for this PR, but worth knowing: the ICU chinese-calendar conversion shared with `Taiwan.php` is off by a day in some years (it returns 2027-02-07 and 2030-02-02 for Lunar New Year's Day, actual dates are 2027-02-06 and 2030-02-03). That's a pre-existing issue, not something you introduced.

freekmurze:
I've pushed a commit to your branch with the fixes, so you don't have to do the round trip. Summary of what changed:

**Ching Ming** is now computed from the Qingming solar term instead of being pinned to 5 April. Fixes 2020, 2021, 2024, 2025, 2028, 2029 and so on.

**Sunday shifting now replaces rather than duplicates.** Sundays are general holidays in Hong Kong in their own right, and the gazette lists only the shifted day, so emitting both produced an extra entry (and the duplicate names on consecutive days). With this change every year from 2020 to 2035 yields exactly 17 general holidays.

**Christmas on a Sunday** now produces 聖誕節後第二個周日 on the 27th instead of reusing the 26th's name, matching how 2022 and 2033 were gazetted.

**Null lunar dates are dropped** instead of being cast to `''`, which Carbon parsed as today.

**Translations**: added the five that were missing, removed 復活節星期一翌日 and 農曆年初五 through 農曆年初十, which can never be produced. There's now a test asserting every generated name has a translation across 2000–2050.

The output matches the gazetted general holidays exactly for 2020 through 2026, which the tests now assert year by year.

One thing I left alone: in 2033 the ICU lunar conversion puts Chung Yeung on 1 October, colliding with National Day. 2033 is the well-known leap-month edge case in the Chinese calendar and it affects `Taiwan.php` equally, so it felt out of scope here.

Thanks again for the contribution, and for the nice Lunar New Year day-naming logic. Merging once CI is green.

freekmurze:
Merged and released in [2.5.0](https://github.com/spatie/holidays/releases/tag/2.5.0). Thanks @littpi!

littpi:
Hello @freekmurze . Thanks for the fix and the comments on helping me acknowledge the issues!

Due to the complexity of Hong Kong Calendar (as it is merging both Western and Chinese holidays), there is an exceptional case which is based on the holiday shifting. In 2026, the holiday name `The day following Easter Monday (復活節星期一翌日)` is conflicted with `The day following Ching Ming Holiday (清明節翌日)`, which is also a shift from the original Ching Ming Holiday. According to their policy, Easter Monday will be shifted to next day. You can check on [our government holiday calendar in 2026](https://www.gov.hk/en/about/abouthk/holiday/2026.htm).

If you don't mind, I can push the new translation again to add the name back to the file.

